PR #84, opened 2014-07-01
Replace chart.js with c3.js

blkperl:
chart.js is poorly maintained and has been disabled. c3.js
has active development and removes the need for two charts
on the facts page.

Co-authored-by: Sage Imel <sage@cat.pdx.edu>
Co-authored-by: Ryan Niebur <ryan@cat.pdx.edu>


nibalizer:
@daenney ping

daenney:
Oh le derp, I keep forgetting this one.

nibalizer:
@blkperl you too. ping ping

blkperl:
@daenney rebased, ready for review

daenney:
I like this.

Only thing is `GRAPH_FACTS`. If someone now defines this in their `settings.py` they'd override the default list we provide which might not be the optimal user experience. Unfortunately the setting loading works by overriding the default setting by a user specified one.

Can we come up with a solution so that we can have both? Something like `GRAPH_FACTS` that is populated by us and a `GRAPH_EXTRA_FACTS` that allows users to expand that list? And then combine those two lists in code before we check whether to graph a fact?

nibalizer:
@daenney there are some hacks around line 30 of app.py that flatten system and user defined GRAPH_FACTS together :)

daenney:
Yeah I know, but that doesn't work. `app.config` is the end representation that's loaded into Flask, it will already have squashed those structures. You can't differentiate between the default and user passed in settings through `app.config`.

nibalizer:
@daenney I just tested it and it works fine. Is this maybe something that works when debug=True but not otherwise?

daenney:
Mmm, that's curious. I'll try a few things.

nibalizer:
@daenney poke

daenney:
I'm still confused as to how this works, but I guess it ends up merging the data structures instead of overriding them.

daenney:
You know what. Fuck it. Lets try this and gather some feedback during the process.

PR #91, opened 2014-09-10
Accept the configuration_version as alternative in the report view.

href:
[puppet-hipchat](https://github.com/jamtur01/puppet-hipchat) has this very helpful feature where the puppet runs that are logged in the chat are posted with a link to the report.

Unfortunately, for now only the '/latest' report can be linked as the report hash is not available in the puppet report format.

I thought the easiest way to change this is to optionally accept the configuration_version as the `<report_id>`. It's a tiny change really, but it would help solve this problem:

https://github.com/jamtur01/puppet-hipchat/pull/23#discussion-diff-10535251

We'll then end up with this:
screenshot of google chrome 10-09-14 16 41 34

### DIFF
--- a/puppetboard/app.py
+++ b/puppetboard/app.py
@@ -226,11 +226,15 @@ def report_latest(node_name):
 def report(node, report_id):
     """Displays a single report including all the events associated with that
     report and their status.
+
+    The report_id may be the puppetdb's report hash or the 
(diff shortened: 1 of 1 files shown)


nibalizer:
+1

daenney:
I don't see how this could ever break anything so...

PR #100, opened 2014-11-24
Initial port of puppet dasboard radiator view

ardichoke:
Ported the puppet dashboard radiator view to puppetboard. This is a simple but useful heads up display of the last run status of all nodes a master knows about.


johnzimm:
I really like the addition of this feature.  I cobbled something together to use in for a "radiator" for sorts and was hoping someone would do a better job of it.

Playing devil's advocate on the color a bit (I know it can be argued either way)....  My current company is a Puppet Enterprise customer and we also use Puppetboard for an unauthenticated read-only view.

In Puppet Enterprise Console (like Puppet Dashboard) we grew accustomed to the "changed" being blue and the "unchanged" being green.  The thinking behind this is that "unchanged" means everything is as it should be.  "changed" means something has happened  and hopefully it was intentional.

In translating this to a monitoring/alerting system (nagios oriented) I would return an "ok/green" if nothing has changed, a "warning/yellow" if something has changed or a noop run wants to apply changes , and a "critical/red" if something fails.  Other states are "unknown/orange".

We also make use of noop in production in some cases.  It is useful for us to know how much of our infrastructure is actually in noop and is in noop and wants to apply changes.  Not really the "puppet way" of doing things, but it's how we've had to implement a few configurations.  If there is accounting for noop that would be awesome in our case.

nibalizer:
What about extending this to the config so users can pick their colors? I would like this feature more than any one color scheme.

ardichoke:
I have actually changed jobs and am no longer using puppetboard, so I'm not the best person to further develop this. As far as I can tell, this contribution has never been merged upstream. If someone would be so kind as to do that, someone else could take over making the requested changes (or not). Either way, I'm no longer the appropriate person to take care of this.

nibalizer:
@ardichoke do you mean the radiator never made it into puppet-dashboard?

johnzimm:
I believe he means this pull request never made it into the main branch.

I have changed jobs as well, but am still using Puppetboard.  I am standing up new puppet server and puppetdb now.  I implemented a radiator and am planning on switching to the implementation in this commit instead.  I'll see about adding the configuration variables for color and update the merge request.

ardichoke:
@johnzimm is correct. I'm sorry if that was not clear from my comment.

mvisonneau:
@nibalizer, any plans about merging this feature ?

nibalizer:
@mvisonneau there was talk of making the colors configurable, which I would greatly prefer. If no one is willing to do this work up-front though, I suppose this can land as is. I'll let this comment sit for a few days and if no one steps forward we can merge.

daenney:
There's a few things that don't super-excite me about this change but not strongly enough to block any of it:

* Configurable colours (as noted by @nibalizer)
* Not a fan of the `<meta content='150' http-equiv='refresh' />`, I'm pretty sure this can be done more elegantly and less invasively by having `/radiator` return JSON instead and then rendering that with some JS that just updates the HTML 
* The amount of CSS, especially the super-specific CSS selectors which are needlessly specific and slow down rendering. For example the `body.radiator_controller table.node_summary tr.unchanged .percent` can be changed to `tr.unchanged .percent` and achieve the same thing (most of the `body.`'s are pointless too as are the `table.node_summary` since almost everything is contained within them).

ardichoke:
I agree the code is less than ideal (side effect of forking code from the Puppet team and integrating it into another project, as well as my own lack of experience in web development, I suppose), but I don't have a puppetboard install to be able to further work on this any more. This either needs to be rejected, or moved upstream and out of my repo, as I am not currently able to do any further work on this and do not believe I will be in a position to do so for the foreseeable future. It would be pointless for people to submit changes to my fork, as I have no means of testing them anyway.

nibalizer:
http://imgur.com/hZGImPY Is what it looks like right now.

ardichoke:
I hate to keep bumping this, but this pull request has been open for over a year at this point. I rebased this commit on the latest version of master. Once again, I don't have an environment available to test, but it doesn't look like anything has changed that would cause this to break. Could someone review it and either pull it in or reject it?  Either way, I'd like to get this wrapped up, as I do not like things lingering about indefinitely.

corey-hammerton:
In ``app.py`` I can see 2 routes for ``/radiator`` and the same function code broken up in 2 places. Can you please clean that up?

daenney:
I think it might be worth redoing this PR. The history is going to be a mess now on account of us not merging this sooner.

ardichoke:
I cleaned up app.py.

If you're going to redo it, then grab a copy of the files and reject the pull request please. I just want this closed out, one way or another.

corey-hammerton:
To keep the commit history clean the code in this PR will have to be redone in a new PR.

PR #108, opened 2014-12-29
Adds custom facts to enumerate in nodes list

vjanelle:
Enumerates a specific fact to show during the node list, which is
configurable as NODES_FACTS in the configuration.

I've added a default entry which is empty, and some detection around if
a fact doesn't exist on a value (which should just print none).

### DIFF
--- a/puppetboard/default_settings.py
+++ b/puppetboard/default_settings.py
@@ -23,3 +23,4 @@ GRAPH_FACTS = ['architecture',
                'osfamily',
                'puppetversion',
                'processorcount']
+NODES_FACTS = []
--- a/puppetboard/templates/index.html
+++ b/puppetboard/templates/index.html
@@ -66,6 +66,9 @@
           <tr>
             <th class="five wide">Status</th>
             <th class="ten wide">Hostname</th>
+            {% for nodes_fact in config.NODES_FACTS -%}
+            <th>{{nodes_fact}}</th>
+            {% endfor -%}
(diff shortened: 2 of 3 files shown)


daenney:
Right. This seems sensible to me. Do you have a screenshot or two, to see how this behaves? I don't have a PuppetDB at hand with sane data to try out your PR right now.

vjanelle:
https://www.evernote.com/shard/s21/sh/ba3bf558-9bfb-442f-bb4f-34708f3eccf2/07afe49672f85385cc9349c8cea523d1/deep/0/Puppetboard.png

vjanelle:
The nodes display panel on the index page has some odd formatting, due to the weights put on the hostname - the two fields I specified get shoved waaaay to the right.

daenney:
Mmm, interesting. I can see this being useful. I'm back at home after the weekend with a functioning PuppetDB, I'll see what I can do about the formatting.

vjanelle:
\o/

Poil:
When I merge this patch I've got an error per node: 
```
 [error] INFO:requests.packages.urllib3.connectionpool:Starting new HTTP connection (1): puppetdb.xxxxxxxxxx.xxxxx
```

daenney:
That's some weird error logging. That message is logged as `INFO` by the python `logging` library, how does it end up with an `[error]` tag on your side?

vjanelle:
I've rebased this so it should be mergeable again.

PR #114, opened 2015-01-15
Fix proxy_pass line in nginx+gunicorn example.

jarpy:
The example was missing the protocol part of the backend server URL
('http://'). This causes nginx to complain of a "malformed URL".


daenney:
Nice catch, thank you!

jarpy:
Oh no, thank you. I'm loving Puppetboard. :)

daenney:
Awesome! Great to hear and thank you for saying that.

PR #117, opened 2015-02-17
Add scroll to top button

juliengk:


daenney:
Could you quickly add a screen cap to this PR? Not as a commit, just drag it into the comment field so we can see what this ends up looking like.

juliengk:
top-button

juliengk:
Let me know if you want to change the size, color...

daenney:
I really have no problems with this, if anything it'll come in useful. We can always tweak the UI later, no need to have the PR hanging any longer.

PR #123, opened 2015-02-27
Fix #77 UI alignment issue

### LINKED ISSUE #77: UI Alignment issue for unreported nodes
The second column in Overview and Nodes pages align poorly when there are unreported nodes. See screenshot
puppetboard

juliengk:


stack72:
@juliengk what UI Alignment issue was being fixed here? Have you got a screenshot of it?

PR #128, opened 2015-03-13
Fix csrf secret key

juliengk:
In the query page i had no result and got the error message "CSRF token missing"


daenney:
I don't think hard-coding a secret key is a good idea. It's why I had the `random` call in there, to make sure that you get a new random one every time. Having a default one in settings will mean people won't override it and we end up with all installations of Puppetboard using the same CSRF token.

juliengk:
csrf_secret A byte string which is the master key by which we encode all values. Set to a sufficiently long string of characters that is difficult to guess or bruteforce (recommended at least 16 characters) for example the output of os.urandom(16). http://wtforms.readthedocs.org/en/latest/csrf.html#module-wtforms.csrf.session

From what i understand that secret key does not change. With app.secret_key set to os.urandom(24), everytime the page is loaded a new secret key is generated.

daenney:
That's true, so that was a flaw. However, you've hardcoded a single secret key now in `default_settings.py` which is what I take issue with. It now means that every single installation of Puppetboard will share that same secret key, which is not a good idea.

juliengk:
then as default we put to None and display a message if they have not set it up

daenney:
Agreed, that seems like a better solution.

juliengk:
change has been made. Let me know what you think

fgimian:
I've modified this pull request to default to a random key but also allow the user to override this with their own statically generated secret key.  This means that those who aren't overly worried about security (most instances of Puppetboard would be deployed internally anyway) won't get hassled with generating a secret key.

Cheers
Fotis

PR #130, opened 2015-03-23
Fix sorting order when computing the largest pieces to included in the p...

sigmunau:
...ie chart

### DIFF
--- a/puppetboard/templates/_macros.html
+++ b/puppetboard/templates/_macros.html
@@ -53,7 +53,7 @@
     value: 0,
     }
   ]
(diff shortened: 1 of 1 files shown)


fgimian:
Thanks for your contribution, though I think this requires some more work as it causes trouble when sorting certain facts:

e.g. before the patch

before

after the patch

after

I'll close this for now and reconsider any further ideas you may have which take all fact types into consideration :smile: 

Cheers
Fotis

sigmunau:
To clarify, this patch sorts the values based on the number of hits. As shown in your example, 14.04 comes first, because it has the most hits, then 12.04 and then 10.04. Without this patch you will get the most common values included in the `other values` chunk and just the least common ones displayed

PR #133, opened 2015-04-29
Feature: adding examples directory

visibilityspots:
I hereby added an examples directory where projects which show off the puppetboard in various manners can be added.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "examples/visibilityspots"]
+	path = examples/visibilityspots
+	url = git@github.com:visibilityspots/vagrant-puppet.git


stack72:
@visibilityspots If I am reading this PR right, this is adding a vagrant playground to see Puppetboard in action - correct?

visibilityspots:
@stack72 yes indeed, I added an examples directory which consists a first submodule pointing to a vagrant playground which is provided by puppet.

nibalizer:
@visibilityspots can you add some README stuff to explain how to use this?

nibalizer:
I would rather just a link in the readme than a submodule.

daenney:
I don't think the whole submodule thing is a great idea either. If it was a dependency that we needed to vendor, sure, but for some examples, no. Little section in the README plus a link to the repository would be enough.

PR #135, opened 2015-05-07
Added inventory page along with a navbar link.

stoyansbg:
I used to update my inventory list automatically via other means, but since I installed PuppetBoard a couple of days ago, I thought it was a good fit. PuppetBoard provides great visibility to the status of the Puppet infrastructure, but I also needed a custom single view of my nodes along with a set of facts. The inventory page usability can be greatly enhanced using custom facts. Some of the custom facts I use with it are "reboot_required", "updates_available", "ufw_active", etc.

Synopsis:
Added inventory page showing a table similar to the "Nodes" one, but containing any customizable set of facts. A setting INVENTORY_FACTS has been added. It controls the list of facts to be displayed along with a description for the header of the table. I would've added a setting to hide the tab altogether if not needed, but it would've required some major changes. I hope you find it useful.

### DIFF
--- a/puppetboard/app.py
+++ b/puppetboard/app.py
@@ -178,6 +178,61 @@ def nodes():
         stream_template('nodes.html', nodes=nodes)))
 
 
+@app.route('/inventory')
+def inventory():
+    """Fetch all (active) nodes from PuppetDB and stream a table displaying
+    those nodes along with a set of facts about them.
+
+    Downside of the streaming aproach is that since we've already sent our
+    headers we can't abort the request if we detect an error. Because of this
+    we'll end up with an empty table instead because of how yield_or_stop
+    works. Once pagination is in place we can change this but we'll need to
+    provide a search feature instead.
+    """
+
+    fact_desc  = []     # a list of fact descriptions to go
+                        # in the table header
+    fact_names = []     # a list of inventory fact names
+    factvalues = {}     # values of the facts for all the nodes
+                        # indexed by node name and fact name
+    nodedata   = {}     # a dictionary containing list of inventoried
+                        # facts indexed by node name
+    nodelist   = set()  # a set of node names
+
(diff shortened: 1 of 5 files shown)


daenney:
What does this look like, screenshot?

stoyansbg:
Here are a couple of screenshots. One with the default settings and one with some custom facts selected. Both images are of the packaged PuppetBoard version 0.0.4, not the current GitHub state although the current does look similar.

inventory-default-settings
inventory-custom-facts

stack72:
@stoyansbg Depending on the facts selected, there could be multiple API calls. Won't this cause extra load on puppetdb?

daenney:
Actually the code does a single call `puppetdb.facts()` so it's not more than one additional query for all the data.

stack72:
cool :+1:

nibalizer:
This is pretty sweet. In testing this @stoyansbg, it looks like the 'filter' text input box isn't the same delicious bootstrappy ness of the filter text boxes on other tabs. Any idea what that is? Easy fix?

stoyansbg:
@nibalizer, the inventory.html template is based on an older version of the nodes.html one prior to the switch to Semantic-UI about a year ago. I used that since it was the one that came from the PyPI packaged 0.0.4 version of puppetboard. I still use it as I like the table view of the old interface better. Anyway, I tested the filtering functionality and it does work for me both in the released(?) 0.0.4 version and the current one. Does the filtering not work for you at all or is it behaving differently? If there's an interest, I can definitely update the template so that it is in line with the current version of nodes.html and the rest of the templates.

stoyansbg:
@nibalizer, the template has been updated and seems to be working as expected. Please try it out and let me know if there's a problem. Thanks!

nibalizer:
Lookin awesome!

PR #136, opened 2015-05-10
Removed footer in the interest of displaying more information

fgimian:
Hey there,

I know this one may be a bit sensitive, but I think that the footer of the site is really not needed as it takes up unnecessary space on the website.

Cheers
Fotis


daenney:
You're right, sensitive and no. Feel free to make it not sticky though so that it's either at the bottom of the viewport if there's no more content or at the bottom of the content.

fgimian:
Sounds fair to me, will do soon :smile:

daenney:
Perfect!

PR #137, opened 2015-05-10
Further details for nodes added (skip count and reported date on overview page)

fgimian:
Hey there again,

I've added a skip count for each node on the overview page along with a reported column which is very helpful on the overview page.

Cheers
Fotis

### DIFF
--- a/puppetboard/static/coffeescript/tables.coffee
+++ b/puppetboard/static/coffeescript/tables.coffee
@@ -13,7 +13,7 @@ $('.facts').tablesorter(
 
 $('.dashboard').tablesorter(
     headers:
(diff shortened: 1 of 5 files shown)


daenney:
Could you squash this down a bit? dfd9007, 51efb12 and 383d6b6 appear to be working on getting the same feature working, in which case I'd rather see that as 1 commit in the history instead of 4.

fgimian:
Done mate, please let me know if that's OK now :smile:

daenney:
You're fast :smile:

fgimian:
Hehe, thanks so much, so are you! :smile:

PR #140, opened 2015-05-10
Patch _macros.html for the latest cleanups

bluecmd:
97bbe96 renamed the keyword arguments for report(node_name, report_id) but _macros.html wasn't updated


bluecmd:
Ping @fgimian

daenney:
Nice catch, ty.

PR #141, opened 2015-05-11
Fixed variabel names in index.html and app.py

ventris:
renamed node in @app.route('/reports/<node>') to node_name
renamed node in {% if node.report_timestamp %} to node_name


fgimian:
Thanks!  I thought I had gotten all of them but missed a couple :smile:

PR #147, opened 2015-06-20
Add /status page for monitoring systems

bluecmd:
This is useful for feeding puppet data into a monitoring system such as Prometheus.

### DIFF
--- a/puppetboard/app.py
+++ b/puppetboard/app.py
@@ -3,6 +3,7 @@ from __future__ import absolute_import
 
 import logging
 import collections
+from dateutil.tz import tzoffset
 try:
     from urllib import unquote
 except ImportError:
@@ -403,3 +404,37 @@ def metric(metric):
         'metric.html',
         name=name,
         metric=sorted(metric.items()))
+
+
+@app.route('/status')
+def status():
+    """Fetch all (active) nodes from PuppetDB and construct a scrapable list.
+
+    This is useful for feeding puppet data into a monitoring system such as
+    Prometheus.
+    """
+    nodelist = puppetdb.nodes(
+          unreported=app.config['UNRESPONSIVE_HOURS'],
+          with_status=True)
+
+    nodes = [x for x in yield_or_stop(nodelist)]
+
+    def generate():
+        for node in nodes:
+            yield 'puppet_node{'
+            yield 'status="%s",' % node.status
+            yield 'node="%s"' % node.name
+            yield '} 1\n'
+        for node in nodes:
+            if node.report_timestamp is None:
+                continue
+            unix_epoch = datetime.fromtimestamp(0, tzoffset('UTC', 0))
+            yield 'puppet_last_updated_timestamp{'
+            yield 'node="%s"' % node.name
+            yield '} '
+            yield str(int((node.report_timestamp-unix_epoch).total_seconds()))
+            yield '\n'
+
+    resp = Response(stream_with_context(generate()))
+    resp.headers['Content-Type'.encode('ISO-8859-1')] = 'text/plain'
+    return resp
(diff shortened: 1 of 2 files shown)


nibalizer:
A feature I've wanted for a long time but haven't implemented yet is to expose content-type json on the normal puppetboard endpoints. Automated systems hitting puppetboard is the future, then puppetboard provides essentially a read-only/filtering proxy to the puppetdb api. I think implementing that would be superior to this, and neatly fit your use case as well. What do you think?

bluecmd:
Prometheus doesn't support JSON, but I'd be all for a ?format=json / ?format=prometheus or something if that's what you mean.

daenney:
So this is currently very much tied to what Prometheus expects so a generic endpoint like `/status` is a bad idea.

As an aside, why the heck doesn't it support a standard like JSON?

bluecmd:
I totally agree that snatching /status and saying it's only available for Prometheus is bad. I do prefer the Prometheus format to inspect as a human however, so I'd say it's a better default, but I don't feel strongly about it.

The reason JSON is bad is that you cannot parse it while streaming it - you need everything. For monitoring data that's especially bad. The preferred format is actually protobuf, but that is a bit more hairy to implement, and is totally not readable by humans.

bluecmd:
To recap: I'd be glad to have /status accept multiple formatters when anyone steps up and wants to write that

nibalizer:
I'm okay with this taking ?format=prometheus  or a Content-Type:Prometheus header or something on the /status route. Sound good? Also is this kind of raw string manipulation the best way right now to generate prometheus data? Is there a pip package for vomiting that stuff that we can require I guess is what I'm asking.

bluecmd:
Fine with me! What would you have the default be? I.e. not having ?format= set.

There is a client library, but since we're not a static daemon the library doesn't do much good. The format is simple enough to allow trivial serialization, so a library doesn't really make much sense in the "collect-and-render" case I'm afraid.

nibalizer:
I think giving a not implemented error, which could be a 200 OK instead of a 501 makes sense if we weren't explicitly asked for format=prometheus.

daenney:
Eh, no. We have HTTP status codes for that. Return 200 OK when we mean "actually we couldn't service this request because the have no clue how to answer for this format" is not alright and a way to break the web.

A 406 "content not acceptable according to the Accept headers" is the correct way to go. This is also what Twitter and other API's return when "an invalid format is specified in the request".

nibalizer:
Ok. (Not breaking the internet)++

jyaworski:
Hello:

Is this still needed? If so, can you rebase and squash?

bluecmd:
Totally forgot this change. We run with this in our production, so we would of course like it to be merged, but with the mentioned modifications above. I haven't gotten around to implement that.

corey-hammerton:
Please also note that Puppetboard is now environment aware and has undergone many changes with the 0.1.0 release. Please update your code to help reflect those changes. 

Might be better to create a new PR

bluecmd:
Since the comments so far have not been about the code but the general design I fail to see how closing this to open a new PR would be better. The whole discussion would be lost. Ultimately though, it's the maintainers' call.

corey-hammerton:
It was mentioned that it would be good to support different content types, i.e. JSON, Prometheus, XML (if possible), etc. That would allow this to be monitored by any number of systems, not just Prometheus.

Even though I'm relatively new when it comes to maintaining this project my experience with application monitoring do not really like template streaming (not to mention that I don't fully understand it myself) so outputting the entire result-set at once is my preference.

bluecmd:
I'm going to abandon this and create a puppet_exporter instead, as I believe that's a better approach for Prometheus and alike. Thanks for the comments all!

PR #156, opened 2015-08-20
puppetboard: Adding Catalog Viewer

corey-hammerton:
This makes use of the available catalog endpoints in pypuppetdb and PuppetDB
and displays to administrators the last compiled catalog for a node,
if the variable ENABLE_CATALOG is True (defaults to False)

This is a re-work of https://github.com/puppet-community/puppetboard/pull/93 and https://github.com/puppet-community/puppetboard/pull/115

### DIFF
--- a/puppetboard/app.py
+++ b/puppetboard/app.py
@@ -403,3 +403,13 @@ def metric(metric):
         'metric.html',
         name=name,
         metric=sorted(metric.items()))
+
+@app.route('/catalog/<node_name>')
+def catalog_node(node_name):
+    """Fetches from PuppetDB the compiled catalog of a given node."""
+    if app.config['ENABLE_CATALOG']:
+        catalog = puppetdb.catalog(node=node_name)
+        return render_template('catalog.html', catalog=catalog)
+    else:
+        log.warn('Access to catalog interface disabled by administrator')
+        abort(403)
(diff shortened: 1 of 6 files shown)


nibalizer:
I've tested this locally and lgtm. @daenney I'll give you a bit of time to poke it then I'll merge.

daenney:
I don't have something useful to test this against so can't do much poking. Use your best judgement :smile:.

janorn:
Is this catalog only supposed to be reachable via the nodes tab?
I doesn't seem to be available from the individual node page.

corey-hammerton:
@janorn This is just an initial implementation of this functionality. There is always room for improvement.

nibalizer:
I agree, I think maybe having a 'Catalogs' at the top and then a per node list might make more sense.

corey-hammerton:
I should be able to put more effort into the catalogs when I'm working on the upgrades required for PuppetDB 3.x. That is going to be my next task on this project.

nibalizer:
Awesome. Also if we could get pagination in pypuppetdb I would be /so/ happy. Also feel free to join #puppetboard and or #puppet-community on freenode to chat with us.

corey-hammerton:
I've started my work on the PuppetDB 3.x work, it's available here: https://github.com/corey-hammerton/puppetboard/tree/puppetdb-3.0

corey-hammerton:
Also: https://github.com/corey-hammerton/puppetboard/commit/30e1cfef4a56968a37daad27c0d13a07b53bf588

PR #157, opened 2015-09-05
puppetboard/app.py: Optimizing the inventory page

corey-hammerton:
Instead of querying the ```/facts``` endpoint in PuppetDB for every fact that it knows about only
query it for the names that we want. The result is a
much smaller result set that is more noticeable on environments with
large numbers of nodes and/or facts.

On one environment with > 2500 fact-names and 35 nodes a curl test
to the facts endpoint without any query string returned in ~3.5 seconds.
Ran another curl test to the facts endpoint with the query string that
the code in this commit creates, the command finished in ~0.040 seconds.

This is an update for the functionality that was introduced in https://github.com/puppet-community/puppetboard/pull/135

### DIFF
--- a/puppetboard/app.py
+++ b/puppetboard/app.py
@@ -201,9 +201,6 @@ def inventory():
                         # facts indexed by node name
     nodelist   = set()  # a set of node names
 
-    # get all the facts from PuppetDB
-    facts = puppetdb.facts()
-
     # load the list of items/facts we want in our inventory
     try:
         inv_facts = app.config['INVENTORY_FACTS']
@@ -219,6 +216,13 @@ def inventory():
     for description,name in inv_facts:
         fact_desc.append(description)
         fact_names.append(name)
+
+    query = '["or", {0}]'.format(
+        ', '.join('["=", "name", "{0}"]'.format(name)
+            for name in fact_names))
+
+    # get all the facts from PuppetDB
+    facts = puppetdb.facts(query=query)
 
     # convert the json in easy to access data structure
     for fact in facts:


corey-hammerton:
On the same environment the inventory page renders in >= 7.5 <= 8.5 seconds. With this code patch applied it renders in 0.4 seconds

daenney:
:tired_face:. There's no good reason why querying bare `/facts` should be slower than querying `/facts` with `?query` containing all the facts. I have no idea why this happens but IMHO this should be raised upstream as a bug.

On a different track, it would be much more beneficial to fix this in pypuppetdb instead of patching around it in Puppetboard.

corey-hammerton:
@daenney This code patch only queries the /facts endpoint for the facts and values that we want. 

As I mentioned in the description (this is ONLY with curl, not pypuppetdb)

On one environment with > 2500 fact-names and 35 nodes a curl test
to the facts endpoint without any query string returned in ~3.5 seconds.
Ran another curl test to the facts endpoint with the query string that
the code in this commit creates, the command finished in ~0.040 seconds.

corey-hammerton:
Curl test without a query string (output redirected to /dev/null):
```
$ time curl localhost:8080/v3/facts > /dev/null

real    0m2.775s
user    0m0.023s
sys     0m0.013s
```

Curl test with a query string (output redirected to /dev/null):
```
$ time curl localhost:8080/v3/facts -XGET -d 'query=["or", ["=", "name", "puppetversion"], ["=", "name", "kernelrelease"], ["=", "name", "fqdn"], ["=", "name", "ipaddress"], ["=", "name", "architecture"], ["=", "name", "operatingsystemrelease"]]' > /dev/null

real    0m0.039s
user    0m0.020s
sys     0m0.004s
```

daenney:
Yeah ok, but this sentence is confusing:

> Instead of querying PuppetDB for every fact that it knows about only querying it for the fact names that it knows about.

I believe that what you meant to say is "only querying it for the fact names that we want to know about", because PuppetDB knows all fact names meaning it would be the same thing as querying a bare `/facts`.

corey-hammerton:
I'll reword it to something like:

Instead of querying the ```/facts``` endpoint in PuppetDB for every fact that it knows about only query it for the names that we want.

corey-hammerton:
Do you think this PR can be merged?

corey-hammerton:
@daenney and @nibalizer, I want to push a 0.0.5 release before we accept the work done for PuppetDB 3.x and Puppet 4. It would be nice to include this as a minor optimization.

nibalizer:
Woo code go fast.

PR #129, opened 2015-03-19
Possibility to list only reports w/ changes

rafaelhenrique:
Hi! This PR is my first PR are not angry with me :/... 
This PR was opened to solve the issue #127 ... review please!

Thanks! ;)

PR #160, opened 2015-09-24
Pr 129

nibalizer:
This is a re-up of #129

### DIFF
--- a/puppetboard/app.py
+++ b/puppetboard/app.py
@@ -184,14 +184,30 @@ def node(node_name):
     node. This includes facts and reports but not Resources as that is too
     heavy to do within a single request.
     """
+    ignore_empty_reports = app.config['IGNORE_EMPTY_REPORTS']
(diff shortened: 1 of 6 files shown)


corey-hammerton:
Giving that we are developing and testing on PuppetDB 3.x systems would it be better if we did any one of the following?

1. Query the reports endpoint where status is not 'unchanged' and noop is false
2. Display report status information like the node status is shown on the Overview and Nodes pages

Also since we're planning on removing the ``limit_reports`` function it might be better to close and re-think this PR

daenney:
Yeah. With pagination in place `limit_reports` needs to go.

corey-hammerton:
``limit_reports`` is removed in https://github.com/puppet-community/puppetboard/pull/176. Along with showing the report status in the node reports table and the reports tab (which is now fully functional)

nibalizer:
Closing

PR #161, opened 2015-09-25
puppetboard: Adding a more intuitive catalog view

corey-hammerton:
A new endpoint in the header, Catalogs, takes the user to a page with a node
table similar to that in the nodes page. This table shows the node with a
link to the node page, the catalog timestamp with a link to the catalog
page and a small form with a select field to be used to compare the
catalog of this row's node with that of another node.

This comes after a discussion with other project maintainers about adding functionality to, in addition to making a Catalogs endpoint, compare the catalogs between 2 nodes. This page's URL looks similar with github's commit/branch compare page with the value /catalogs/compare/<compare>...<against>

This has been tested against a PuppetDB 3.x system with 2 nodes but it uses the query functionality of the catalogs endpoint that is available in PuppetDB 2.x systems.

### DIFF
--- a/puppetboard/forms.py
+++ b/puppetboard/forms.py
@@ -25,4 +25,4 @@ class QueryForm(Form):
 class CatalogForm(Form):
     """The form used to compare the catalogs of different nodes."""
     compare = HiddenField('compare')
-    against = SelectField(u'against')
+    against = SelectField('against')
--- a/puppetboard/forms.py
+++ b/puppetboard/forms.py
@@ -2,7 +2,10 @@ from __future__ import unicode_literals
 from __future__ import absolute_import
 
 from flask.ext.wtf import Form
-from wtforms import RadioField, TextAreaField, validators
+from wtforms import (
+    HiddenField, RadioField, SelectField, 
+    TextAreaField, validators
+)
 
 
 class QueryForm(Form):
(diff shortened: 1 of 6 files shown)


corey-hammerton:
Does anybody like this idea or have had a chance to test it out?

I would like additional testing before I merge it into the main tree. This is intended to be included in the 0.1.0 release.

daenney:
It would be great if we could get a couple of screenshots of what this is like and how it's to be used. Picture is worth a thousands words, especially when it comes to a web interface.

puppetboard/forms.py
```diff
@@ -18,3 +21,8 @@ class QueryForm(Form):
         ('reports', 'Reports'),
         ('events', 'Events'),
         ])
+
+class CatalogForm(Form):
+    """The form used to compare the catalogs of different nodes."""
+    compare = HiddenField('compare')
+    against = SelectField(u'against')
```
daenney: Be consistent with your unicode declarations please.

corey-hammerton:
The view of the Catalogs Endpoint:
catalogs

The view of the Catalog Compare page:
catalog_compare

corey-hammerton:
If there are no objections I see no reason not to merge this PR

PR #164, opened 2015-10-06
Make Puppetboard work with the PuppetDB V4 api

raphink:
This PR adapts puppetboard to work with https://github.com/puppet-community/pypuppetdb/pull/58

All the features seemed to work fine with that.

PR #163, opened 2015-10-05
Use more standard colors for Puppet run states

raphink:
Other tools such as Dashboard and Foreman use red for failed, green for unchanged and blue for changed resources. This PR adjusts the colors used so that Puppetboard behaves like the rest of these interfaces.

Result:

Overview
Nodes view
Report events view

PR #166, opened 2015-10-07
Include report status and logs

raphink:
This embarks both #164 and #163.

Here are some screenshots:

### Overview with striped status

Overview with striped status

### Nodes list with striped status

Nodes list with striped status

### Logs in the report view:

Notice logs

Failed logs (truncated)

The last log screenshot is truncated as it requires scrolling to be viewed.

### DIFF
--- a/puppetboard/app.py
+++ b/puppetboard/app.py
@@ -130,6 +130,7 @@ def index():
         'noop': 0
         }
 
+    report_status = {}
     for node in nodes:
         if node.status == 'unreported':
             stats['unreported'] += 1
@@ -141,15 +142,21 @@ def index():
             stats['noop'] += 1
         else:
             stats['unchanged'] += 1
+        reports = get_or_abort(puppetdb._query, 'reports',
+                               query='["=","certname","{0}"]'.format(node.name),
+                               limit=1)
+        if len(reports) > 0:
+            report_status[node.name] = reports[0]['status']
 
-        if node.status != 'unchanged':
+        if node.status != 'unchanged' or report_status[node.name] == 'failed':
             nodes_overview.append(node)
 
     return render_template(
         'index.html',
         metrics=metrics,
         nodes=nodes_overview,
-        stats=stats
+        stats=stats,
+        report_status=report_status
         )
 
 
@@ -169,14 +176,20 @@ def nodes():
         unreported=app.config['UNRESPONSIVE_HOURS'],
         with_status=True)
     nodes = []
+    report_status = {}
     for node in yield_or_stop(nodelist):
         if status_arg:
             if node.status == status_arg:
                 nodes.append(node)
         else:
             nodes.append(node)
+        reports = get_or_abort(puppetdb._query, 'reports',
+                               query='["=","certname","{0}"]'.format(node.name),
+                               limit=1)
+        if len(reports) > 0:
+            report_status[node.name] = reports[0]['status']
     return Response(stream_with_context(
-        stream_template('nodes.html', nodes=nodes)))
+        stream_template('nodes.html', nodes=nodes, report_status=report_status)))
 
 
 @app.route('/inventory')
--- a/puppetboard/app.py
+++ b/puppetboard/app.py
@@ -323,12 +323,13 @@ def report(node_name, report_id):
 
     for report in reports:
         if report.hash_ == report_id or report.version == report_id:
-            events = puppetdb.events(query='["=", "report", "{0}"]'.format(
-                report.hash_))
+            events = report.events()
+            logs = iter(report.logs)
             return render_template(
                 'report.html',
                 report=report,
-                events=yield_or_stop(events))
+                events=yield_or_stop(events),
+                logs=yield_or_stop(logs))
     else:
         abort(404)
 
--- a/puppetboard/static/css/puppetboard.css
+++ b/puppetboard/static/css/puppetboard.css
@@ -36,6 +36,50 @@ th.tablesorter-headerDesc::after {
   display: block;
 }
 
+.ui.header.failed, .ui.line.failed {
+  color: #AA4643;
+}
+
+.ui.label.failed {
+  background-color: #AA4643;
+}
+
+.ui.header.changed, .ui.line.changed {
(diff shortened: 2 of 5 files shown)


puppetboard/app.py
```diff
@@ -142,15 +142,21 @@ def index():
             stats['noop'] += 1
         else:
             stats['unchanged'] += 1
+        reports = get_or_abort(puppetdb._query, 'reports',
+                               query='["=","certname","{0}"]'.format(node.name),
+                               limit=1)
+        if len(reports) > 0:
+            report_status[node.name] = reports[0]['status']
```
corey-hammerton: The Reports endpoint in PuppetDB 3.0 and above support the latest_report? query field. Please modify your query to look something like this (Can remove the limit argument too):

'["and", ["=", "certname", "{0}"], ["=", "latest_report?", true]]'

puppetboard/app.py
```diff
@@ -170,14 +176,20 @@ def nodes():
...
         if status_arg:
             if node.status == status_arg:
                 nodes.append(node)
         else:
             nodes.append(node)
+        reports = get_or_abort(puppetdb._query, 'reports',
+                               query='["=","certname","{0}"]'.format(node.name),
+                               limit=1)
+        if len(reports) > 0:
+            report_status[node.name] = reports[0]['status']
```
corey-hammerton: Please see note above regarding the latest_report? query field in the Reports endpoint.

requirements.txt
```diff
@@ -5,5 +5,5 @@ MarkupSafe==0.19
 WTForms==1.0.5
 Werkzeug==0.9.4
 itsdangerous==0.23
-pypuppetdb==0.1.1
+git+https://github.com/raphink/pypuppetdb@v4-api#egg=pypuppetdb-0.1.0
```
corey-hammerton: Pypuppetdb 0.1.0 has been released on PyPi

daenney:
I'm not at all a fan of the striped colours. I find it really hard to read any of the text. I also don't see the need for Compilation Failed in two varieties since the number next to it indicating the amount of changes, successful and failed will already tell you what happened.

puppetboard/app.py
raphink: Ah great, I guess it also works better.

requirements.txt
raphink: Yes, I have another pr to fix for that and then I'll rebase. @mcanevet do you think you can fix the branch maybe?

corey-hammerton:
I agree with @daenney on this. I am not a fan of the striped colours

raphink:
OK. How about an additional info on the side (say, a very thin column) to add that info?

daenney:
What info do you feel is missing? We have counts for successful, skipped and failed resources.

raphink:
Whether the catalogue compiled.

daenney:
If the catalogue failed to compile there can be no changes to the machine so how did you end up with the Catalogue Failed (changed) state?

duritong:
Normal puppet runs just fall back to the previously cached catalog, apply it and then report against that one.

This is something that is super annoying with the puppet-dashboard, as the only way you realize failing catalog compiles is by either looking at logs or comparing the reported catalog versions.

Having an indication that the compile fails would really, as it usually points to something broken that should be fixed asap.

corey-hammerton:
PuppetDB only stores the most current working catalog so comparing between catalog versions is impossible. 

Furthermore since puppet doesn't report to PuppetDB that a catalog compilation has failed one of the only ways that I see PuppetBoard users deducing this outcome is comparing the catalog timestamp with the latest report timestamp. This will be tricky to program.

propyless:
Hi...
@corey-hammerton PuppetDB since 3.x stores if the catalog compilation failed. PuppetDB 3.x also stores the agent run report.

so while PanoPuppet for puppetdb 2.x does what you proposed, compare catalog timestamp with the latest report timestamp (it was not tricky to program)
Now instead for puppetdb 3.x, PanoPuppet checks the latest report hash for each node.
You can see here: https://docs.puppetlabs.com/puppetdb/3.1/api/query/v4/reports.html
there is a new key in the hash for each report:
  "status": <status of node after report's associated puppet run>,
which can either be failed, pending or success.
So you are in fact able to deduce from the status of a report and lack of events for a report to see if its a failed compilation.
Either that or you can parse the agent log found here: GET /pdb/query/v4/reports/<HASH>/logs

raphink:
@daenney like @duritong said: compilation failed but cached catalogue caused changes.

raphink:
Hence my initial striped colours proposal.

corey-hammerton:
Not necessarily, if only manifest changes were applied to the puppet master.

raphink:
I'm not sure to understand.

corey-hammerton:
Nevermind then. 

Have you tested with PuppetDB 3.x, PyPuppetDB 0.2.0 and a working version of PuppetBoard with these changes?

raphink:
Not yet, and I'm currently afk until around the 10th of November.

nibalizer:
What is the status of this? ping @raphink

jyaworski:
Hello:

Is this still needed? If so, can you rebase and squash?

jamgregory:
Just my two pence: I think it'd be good to be able to see logs in Puppetboard. 

I'm currently using Puppet Dashboard, and I do use the logs from time to time to see what exactly failed (not always easy to see when running the Puppet agent with -t, as the errors can sometimes be obscured by other messages). Would it be possible to get that functionality added please?

daenney:
@jamgregory That's a different thing, please raise an issue about that.

jamgregory:
@daenney: oh? Judging by the last screenshot in the PR, that was what I was thinking? Happy to open a separate issue if it isn't.

daenney:
Uh, sorry. Getting things mixed up.

jamgregory:
No worries @daenney :+1:

mcanevet:
@raphink could you please rebase this?

corey-hammerton:
Closing if favour of, the now merged, https://github.com/voxpupuli/puppetboard/pull/163

Issue #165, opened 2015-10-07
ArchLinux package is gone


tuxinaut:
The ArchLinux package **Maintained by Niels Abspoel** is gone the link responds with a 404.

daenney:
What do you expect as an outcome? We currently don't maintain that package :smile:

bastelfreak:
@tuxinaut I added [python2-puppetboard](https://aur.archlinux.org/packages/python2-puppetboard/), [python-pypuppetdb and python2-pypuppetdb](https://aur.archlinux.org/packages/?O=0&SeB=nd&K=puppetdb&outdated=&SB=n&SO=a&PP=50&do_Search=Go) to the AUR4 and will shortly move it into the official community repo.

tuxinaut:
Thank you for be the new maintainer for this package :+1:

PR #167, opened 2015-10-08
Remove unmaintained Archlinux package entry

tuxinaut:
Remove unmaintained Archlinux package entry fix for #165


daenney:
Closing in favour of #175

PR #168, opened 2015-10-09
Introducing simple plugin system for Puppetboard

marcinzaremba:


corey-hammerton:
What does this PR hope to achieve for PuppetBoard?

daenney:
I guess the idea is that people can then extend Puppetboard with custom apps and theoretically we can split up the core and make everything an app. I've toyed with the idea before and it's probably a nice way forward however setuptools entry points would be a better way to go about this for discovering 'plugins' than a custom loading mechanism.

nibalizer:
I'd like to see us have an example plugin (even if it is trivial), that way we have confidence that the plugin system is working and plugin authors have an example.

divad:
Shouldn't you use http://flask.pocoo.org/docs/0.10/blueprints/ (Blueprints in Flask) to do this? Seems like the right mechanism for extending a Flask app...

daenney:
@divad Take a closer look at the code, the suggested solution registers the "plugins" as blueprints:

```python
app.register_blueprint(module.app.main, url_prefix=prefix)
```

jyaworski:
Hello:

Is this still needed? If so, can you rebase?

mterzo:
Going to close this in favor of #322

PR #170, opened 2015-10-12
Make "skips" color consistent

rfletcher:
"Skips" were highlighted with orange on one page and yellow on another. This change makes them consistent, and switches the color to "black" for accessibility reasons. (The contrast between the shades of orange and red used for skips and errors was *very* low. Indistinguishable on some screens.)

### DIFF
--- a/puppetboard/static/css/puppetboard.css
+++ b/puppetboard/static/css/puppetboard.css
@@ -75,6 +75,23 @@ th.tablesorter-headerDesc::after {
   color: #FFF;
 }
 
+.ui.menu.yellow {
+  background-color: #FFFF64;
+}
+
+.ui.yellow.header, i.yellow {
+  color: #FFFF64;
+}
+
+.ui.labels .yellow.label::before, .ui.yellow.labels .label::before, .ui.yellow.label::before {
+  background-color: #FFFF64;
+}
+
+.ui.yellow.labels .label, .ui.yellow.label {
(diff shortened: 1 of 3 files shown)


corey-hammerton:
Skips should be yellow or orange, black is already used by unreported nodes.

daenney:
If contrast is an issue then we should fix that but yellow/orange is recognised as a warning state/colour by most human beings. Black is associated with death usually so a better fit for unresponsive/decommissioned nodes.

rfletcher:
I agree about yellow being a more traditional "warning" color. I didn't use that because the bundled UI framework doesn't define a "yellow" in its CSS. (The "yellow" class I replaced above does nothing.)

I see now that Puppetboard [already defines a "darkblue" color](https://github.com/puppet-community/puppetboard/blob/master/puppetboard/static/css/puppetboard.css#L68-L76) in addition to those provided by the framework. I'll run with that and define a "yellow".

rfletcher:
Ok, now using the "yellow" class in both places, and yellow exists. It'll look something like this:

screen shot 2015-10-12 at 8 22 45 pm

daenney:
Can we tone down the yellow just a bit? It's very bright compared to the green and the red.

rfletcher:
screen-shot-2015-10-12-at-8 22 45-pm

daenney:
Much better!

PR #173, opened 2015-10-23
update query example to use certname

pyther:
The query field `name` was replaced with `certname`. https://docs.puppetlabs.com/puppetdb/3.1/api/query/v4/nodes.html#query-fields

`["=", "name", "hostname"]` fails where as `["=", "certname", "hostname"]` works as expected


corey-hammerton:
Good catch. I'll include this in the updates for the PuppetBoard release with PuppetDB 3.x support.

PR #182, opened 2015-11-11
puppetboard/templates/reports: Passing current_env to the reports_table macro

corey-hammerton:
This fixes https://github.com/puppet-community/puppetboard/issues/181


gotyaoi:
I just tested out this patch and it works for me.

PR #183, opened 2015-11-13
version number updated in templates/layout.html

ghakfoort:
the version of puppetboard was updated to 0.1.1 but the version number in this layout file was still 0.1.0.


corey-hammerton:
I was hasty in release 0.1.1 the other day in order to release the reports fix. 

I will merge this when I get a chance but will not release another package containing only this fix.

PR #185, opened 2015-11-16
Allow to list all nodes again (remove environment filter)

raphink:
Implies and includes https://github.com/puppet-community/puppetboard/pull/184


mcanevet:
+1

nibalizer:
:+1:

puppetboard/app.py
```diff
@@ -362,16 +384,22 @@ def reports(env, page):
     """
     check_env(env)
 
+    if env == '*':
+        reports_query = None
+        total_query = '["extract", [["function", "count"]]]'
```
corey-hammerton: Only PuppetDB 3.2 supports the count function WITHOUT a query expression. My first implementation of this was like this:

``'["extract", [["function", "count"]], ["~", "certname", ""]]'``

raphink: OK, and this one is backward compatible I suppose

corey-hammerton: This patch that I suggested will be compatible with PuppetDB 3.0 and 3.1. No need to break it.

PR #187, opened 2015-11-18
Add wsgi.py and Dockerfile

raphink:
This PR adds support for easily developing and deploying Puppetboard with Docker.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,3 @@
+FROM grahamdumpleton/mod-wsgi-docker:python-2.7-onbuild
+CMD [ "wsgi.py" ]
+
--- a/puppetboard/default_settings.py
+++ b/puppetboard/default_settings.py
@@ -1,6 +1,6 @@
 import os
 
-PUPPETDB_HOST = 'localhost'
+PUPPETDB_HOST = 'puppetdb'
 PUPPETDB_PORT = 8080
 PUPPETDB_SSL_VERIFY = True
 PUPPETDB_KEY = None
--- a/wsgi.py
+++ b/wsgi.py
@@ -1,0 +1,3 @@
+from __future__ import absolute_import
+import os
+from puppetboard.app import app as application


corey-hammerton:
I apologize for the late reply. Have you tested this new Docker file? I currently don't have a available resource for test a docker image.

raphink:
We use puppetboard with docker in production, so yes, this is tested.

corey-hammerton:
Thanks @raphink for the clarification. Since this implies that a new package system must be maintained I've added this PR to the 0.2.0 milestone, also because this involves changing an existing value in ``default_settings.py``

corey-hammerton:
Is there a way to provide an external configuration option for their docker containers, for example provide a ``settings.py`` file to be accessible via the ``env['PUPPETBOARD_SETTINGS']``?

corey-hammerton:
@raphink I will merge this into the tree. However there will not be an officially maintained Docker image. I have found nothing on Docker Hub that suggests the voxpupuli community maintains any images.

raphink:
@corey-hammerton there's no images yet indeed. @puppetlabs has started maintaining images though, and I think it would make sense to have an account for @voxpupuli too.

corey-hammerton:
One last thing. I'm not comfortable with changing the `PUPPETDB_HOST` value in `default_settings.py` because it might cause unnecessary disruption for users upgrading to new versions.

raphink:
I could add a script to edit that value when building the image, like we do in ours. `localhost` makes no sense in a container.

corey-hammerton:
Or, probably preferably, we could maintain a separate branch that will contain docker specific files/settings. Then use that for building and maintaining the images.

bastelfreak:
I think a deploy script that does the changes via travis or $awesomething is the better approach compared to a seperate branch. Two branches will result in two sources of truth - which is bad.

raphink:
Yes, two branches will cause issues.

corey-hammerton:
I will merge this, however, until we figure out how we can configure puppetboard settings in the image there will not be an officially supported Docker image. 

I'll also add to the README that this is planned.

PR #195, opened 2015-11-30
Fix index in format() call

raphink:


corey-hammerton:
Duplicates https://github.com/puppet-community/puppetboard/pull/191

PR #196, opened 2015-12-01
Add OVERVIEW_FILTER

raphink:
This allows to filter out nodes in the overview by passing queries to the PuppetDB in the `OVERVIEW_FILTER` configuration parameter.

For example, if I don't want to see the machines whose certname matches `wrk`, I can use:

```
OVERVIEW_FILTER = '["not", ["~", "certname", "wrk"]]'
```

### DIFF
--- a/puppetboard/app.py
+++ b/puppetboard/app.py
@@ -140,7 +140,7 @@ def index(env):
     check_env(env, envs)
 
     if env == '*':
-        query = None
+        query = app.config['OVERVIEW_FILTER']
 
         prefix = 'puppetlabs.puppetdb.query.population'
         num_nodes = get_or_abort(
@@ -157,9 +157,11 @@ def index(env):
         metrics['avg_resources_node'] = "{0:10.0f}".format(
             avg_resources_node['Value'])
     else:
-        query = '["and", {0}]'.format(
-                ", ".join('["=", "{0}", "{1}"]'.format(field, env)
-                    for field in ['catalog_environment', 'facts_environment']))
+        conditions = ", ".join('["=", "{0}", "{1}"]'.format(field, env)
+                  for field in ['catalog_environment', 'facts_environment'])
+        if app.config['OVERVIEW_FILTER'] != None:
+            conditions += ", {0}".format(app.config['OVERVIEW_FILTER'])
+        query = '["and", {0}]'.format(conditions)
         num_nodes = get_or_abort(
             puppetdb._query,
             'nodes',
--- a/puppetboard/default_settings.py
+++ b/puppetboard/default_settings.py
@@ -18,6 +18,7 @@ LOGLEVEL = 'info'
 REPORTS_COUNT = 10
 OFFLINE_MODE = False
 ENABLE_CATALOG = False
+OVERVIEW_FILTER = None
 GRAPH_FACTS = ['architecture',
                'clientversion',
                'domain',


mcanevet:
+1

nibalizer:
:+1:

nibalizer:
@corey-hammerton ping?

corey-hammerton:
From a coding standpoint this patch looks good. But from a management/user perspective we would effectively be allowing administrators to mask a potential problem in their infrastructure. 

I'm speaking from my experience as a system administrator. 

To what benefit would this patch serve?

raphink:
Here's a use case: we manage laptops that are often of the grid. They are thus usually reported on the main view and pollute it, because we don't mind. If I want to see them, I just need to go to the nodes tab where they are not filtered out.

nibalizer:
@raphink I think its really cool that your laptops are managed.

Maybe what we need is multiple overview pages. I could see something in the config file where you specify an array of queries, and each query is used to generate a different overview page.

mcanevet:
@corey-hammerton @nibalizer what prevent this PR to be merged?

corey-hammerton:
I agree with @nibalizer 's last comment. This would be good if your Puppet manages laptops but not server environments. Better off having specific overview pages with their own queries than to add potentially hindering filtering on the general overview page (don't want users, intentionally or not, masking problems in their Puppet infrastructure when managing server environments.

raphink:
Just rebased that. I agree that adding the ability to have multiple overview pages would be nice, but I think this patch to filter out machines from the overview is rather simple and non-obtrusive still.

puppetboard/app.py
```diff
@@ -549,7 +551,9 @@ def report_latest(env, node_name):
     check_env(env, envs)
 
     if env == '*':
-        node_query = '["=", "certname", "{0}"]'.format(node_name)
+        query='["and",' \
+            '["=", "certname", "{0}"],' \
+            '["=", "latest_report?", true]]'.format(node_name)
```
corey-hammerton: We cannot accept this PR with the current changes to the ``report_latest()`` function. Please keep it just to the Overview

corey-hammerton:
Is it at all possible to review your Puppet environment setup? If you're managing laptops as well as other machines and they're on different environments would the environment filters achieve the same thing?

raphink:
@corey-hammerton in our case, we use environments for the code workflow. We have feature environments, as well as "staging" and "stable", so code flows from feature to staging to stable. Some non-critical machines are tagged to always use "staging", while critical machines use the "stable" environment. For that reason, we want "*" to be the default environment in the overview (otherwise, we never see all the machines).

corey-hammerton:
After much thought and consideration I've decided to close this PR.

With the reasons above it seems to me that this code change duplicates already existing functionality with environment filters. In addition, code maintainability in the long run may become a more difficult task when building the queries for the overview page. 

Thank you for your submission.

raphink:
While I can understand your reasoning, this is a feature we use in production and I'd required by some of our clients. This means I will have to maintain a fork with that patch over time.

I would really appreciate if you could suggest an alternate implementation that you would be happy to merge, as maintaining a fork is really time-consuming.

nibalizer:
Seems like long term puppetboard could end up looking like grafana, with queries saved as browsable dashboards. I thought we'd sortof agreed user-defined dashboards was ok?

corey-hammerton:
I will consider this for the 0.2.0 release.

mcanevet:
\o/

raphink:
Ah great news

raphink:
@corey-hammerton what's the status for this?

corey-hammerton:
Will merge when I'm able to.

raphink:
Thanks

mcanevet:
\o/

PR #262, opened 2016-06-28
Feature: Limit nodes report list to failed and changed status

mterzo:
I've been using puppetboard for a while, and I want to see when nodes where changed rather than seeing that my nodes have been unchanged for the last 10 when inspecting the node.

I thought about consolidating the unchanged into counts and time ranges, but then realized that would make the shared macro for displaying reports more complicated and it was already used on the /nodes route.  This is what I whipped up in 15 minutes.


corey-hammerton:
Thank you for your submission, however, I have just created #264 to add more flexible query clauses to the PuppetDB queries for the Nodes and Reports tab based on the value of an optional `status` HTTP query string argument.

PR #268, opened 2016-07-13
Fix radiator view auto-resize

amuhametov:
Radiator view is broken in 0.2.0 due to absence of /static/js/jquery.min.js 
This patch creates symlink to required /puppetboard/static/jquery-2.1.1/jquery.min.js in puppetboard/static/js

### DIFF
--- a/puppetboard/static/js/jquery.min.js
+++ b/puppetboard/static/js/jquery.min.js
@@ -1,0 +1,1 @@
+../jquery-2.1.1/jquery.min.js


mterzo:
+1

corey-hammerton:
Thank you @amuhametov. However, I am refraining from adding symlinks to this project. Can you please update `radiator.html` with the new jquery.min.js URL in this PR?

mterzo:
https://github.com/voxpupuli/puppetboard/blob/master/puppetboard/static/js/jquery.min.map

Should this be changed as well then?

corey-hammerton:
Yes please.

mterzo:
Should we also rename the directory to just jquery. Probably should consider moving to a newer version at some point.

corey-hammerton:
We use other frameworks whose versions are in their directory names, SemanticUI for instance. It's the most obvious way for developers to know which versions we are using. In my opinion it is better to keep the directory names as is.

Upgrading jquery is in my bucket list which will be done in a future PR.

mterzo:
I created 2 issues that this PR brought up. I have 2 change sets that I'll submit later for review.

corey-hammerton:
Closing in favour of #272

Issue #83, opened 2014-06-26
Structured facts display could be improved


gertvdijk:
[Facter 2.0 introduced structured facts](http://docs.puppetlabs.com/facter/2.0/release_notes.html) (compared to only key-value pairs in 1.x). In the 2.0.x series of Facter, no structured facts were provided by default. However, as stated in the [Facter 2.1.0 announcement](https://groups.google.com/d/msg/puppet-users/sbwKLwLa4r0/_Xab9lmLfroJ), now `partitions` is the first standard structured fact (hash).

Please add support for this in Puppetboard. By support I mean one should think of a way it can be displayed 'nicely', rather than for example:

```
partitions  {"vda1"=>{"uuid"=>"af740304-b2c6-440f-877a-5fb15cf86c39", "size"=>"12578816", "filesystem"=>"ext4"}}
```

corey-hammerton:
Starting with PuppetDB 3.0 Structured facts are available in JSON format and new endpoints fact-paths, fact-contents and factsets are available. 

As @daenney said it is a lot of work to implement reliable browsing in PuppetBoard with both string and JSON type fact values. This will be on the backlog but if anybody really needs this support please submit a PR with a link to this issue.

daenney:
So Puppetboard has a `jsonprint` template filter that you can use, see `templates/metric.html` and `templates/metric.html`.

It doesn't provide for a nice way to browse and click through it but will at least print it as more humanly readable JSON instead of a long string.

corey-hammerton:
We are open to PRs if anyone wants to attempt to implement an API for this. Hopefully it can make it into the 0.2.x series.

corey-hammerton:
As of PuppetDB 4.2.0 a new Inventory endpoint is being provided as well as "dotted query" syntax is made available, also the dot notation is experimental.

It may be possible to make use of these new features in the 0.3.0 series with PuppetDB 5.0.0.

gdubicki:
Fixed in v3.3.0 released today. :)

PR #279, opened 2016-07-16
Add a setting to hide fact values from the node page.

TheFlyingCorpse:
Allows the use of a new setting to hide fact values from the node page. This is quite useful on facts which return large results.
Example is the one fact which returns a list of services installed in the OS with some meta information polled from each service.

### DIFF
--- a/puppetboard/templates/_macros.html
+++ b/puppetboard/templates/_macros.html
@@ -1,4 +1,4 @@
-{% macro facts_table(facts, current_env, autofocus=False, condensed=False, show_node=False, show_value=True, link_facts=False, margin_top=20, margin_bottom=20) -%}
+{% macro facts_table(facts, current_env, autofocus=False, condensed=False, show_node=False, show_value=True, link_facts=False, margin_top=20, margin_bottom=20, hidden_facts_values=[]) -%}
 <div class="ui fluid icon input hide" style="margin-bottom:20px">
   <input {% if autofocus %} autofocus="autofocus" {% endif %} class="filter-table" placeholder="Type here to filter...">
 </div>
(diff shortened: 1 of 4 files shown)


puppetboard/templates/_macros.html
```diff
@@ -25,17 +25,21 @@
       {% endif %}
       {% if show_value %}
       <td style="word-wrap:break-word">
-        {% if link_facts %}
-          {% if fact.value is mapping %}
-            <a href="{{url_for('fact_value', env=current_env, fact=fact.name, value=fact.value)}}"><pre>{{fact.value|jsonprint}}</pre></a>
+        {% if fact.name in hidden_facts_values %}
+          <a>HIDDEN</pre>
```
mterzo: Looks like you are missing the closing 'a'

TheFlyingCorpse:
Thanks for pointing that one out. I clearly missed one. It is fixed now!

mterzo:
Actually turning this into a hide/show option. Moment you turn it off you are gonna want to see that value.

TheFlyingCorpse:
The issue/PR for this is the case that Chrome will freeze for whatever amount of time it takes to load all the data. I have 500+ servers what displays a string (json content) of 30k chars+ in a single fact alone. There are a few other facts coming which is going to end up around 1-10k chars that also isn't to display its values in Puppetboard, just that the fact is present from the server in the node page.

corey-hammerton:
If this is a JSON fact then I'll draw your attention to #83. 

This does, however, sound like the Chrome problem described may be something also applicable to other browsers too. Common when trying to load large assets or pages. I'm not sold on the practical uses of this PR.

corey-hammerton:
Another suggestion:

Since this HIDE_FACTS_VALUES_FROM_NODE_VIEW exists in `default_settings.py` you can make direct references to it in the `facts_table` macro without having to instantiate it in `app.py` and calling the macro in `node.html`

For example:

`
{% if fact.name in config.HIDE_FACTS_VALUES_FROM_NODE_VIEW %}
  <pre>Hidden</pre>
{% else %}
  ...
{% endif %}
`

juniorsysadmin:
@TheFlyingCorpse Closing due to inaction, feel free to re-open

PR #284, opened 2016-07-27
Travis: run tests under python 3.5

nikolaik:
Looks like things run well under python 3.5 on our Ubuntu 16.04 install with PuppetDB 4.x. Run tests on python 3.5 to make sure things do not break.


mterzo:
Thanks. I had this in my branch and never submitted the PR. +1

PR #271, opened 2016-07-14
puppetboard/app.py: Enhancing queries for Node and Report states

corey-hammerton:
This resolves #264

On the Nodes and Reports tabs when the user adds a status query string
argument additional query clauses are generated based on its value.
Can be one of failed, changed, unchanged, noop or unreported (for Nodes
only)

No query clause is generated for noop on the Nodes tab. The query field
latest_report_noop was added in PuppetDB 4.1 and we do not want to break
compatability between minor or bug-fix versions.

PR #285, opened 2016-07-27
Pagination prototype

redref:
Hi,

I recently installed puppetboard, but was not able to go in production. With thousands hosts, I got some issues with response time and laggy browser.

I am trying to patch in order to get pagination everywhere (PR code). But, I'm kinda stuck with pypuppetdb nodes endpoint. I can not find a reliable way to count nodes on status.

Looking at pypuppetdb, this situation is normal as status is determined only after request. Do you think (in a reasonable future) every field can / may be suitable for filter or count ?

I am available to work on it if this is ok for you.

### DIFF
--- a/puppetboard/app.py
+++ b/puppetboard/app.py
@@ -7,8 +7,9 @@ try:
     from urllib import unquote
 except ImportError:
     from urllib.parse import unquote
-from datetime import datetime
+from datetime import datetime, timedelta
 from itertools import tee
+from copy import copy
 
 from flask import (
     Flask, render_template, abort, url_for,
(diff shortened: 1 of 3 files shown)


corey-hammerton:
Thank you for your submission. I do, however, have some notes for this PR in no particular order.

- Please restrict your changes to `default_settings.py` to ONLY the entries you wish to add. This includes whitespace
- Please especially refrain from changing the values of existing configuration settings, the PR may be closed as a result if the changes are not reverted.
- The development work done in #271 can be implemented for the Nodes page for pagination and result limiting
- The PuppetDB queries for the index page can be improved to exclude nodes in an 'unchanged' state, will create a separate issue for this
- The index page will only provide users with an overview to their environment's health, any functionality about result limiting and pagination should be done in the nodes page (see Point 4)

PR #288, opened 2016-07-28
Add support for containerized PuppetBoard

tux-o-matic:
The current Dockerfile in the project doesn't deliver PuppetBoard as a ready to use container.
Some changes to the code were also necessary in order to support specifying PuppetDB connection parameters to an immutable container image via ENV parameters set at runtime.
- The most important PuppetDB connection info will be first looks up in current ENV and fallback to configuration file.
- A new route was added to serve static resources directly via Gunicorn/Flask. This way it's not necessary to add a reverse proxy in front of the Puppet Board instance just to serve /static.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,3 @@
+[report]
+exclude_lines =
+    pragma: notest
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,8 @@
 *.py[cod]
 
+# Editor tmp files
+.*.sw*
+
 # C extensions
 *.so
 
@@ -22,6 +25,7 @@ lib64
 pip-log.txt
 
 # Unit test / coverage reports
+.cache
 .coverage
 .tox
 nosetests.xml
@@ -36,3 +40,6 @@ nosetests.xml
 
 # PuppetDB Settings
 /settings.py
+
+# Virtual Environment
+venv
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,13 @@
+language: python
+
+python:
+  - "2.6"
+  - "2.7"
+install:
+  - pip install -r requirements.txt
+  - pip install -r requirements-test.txt
+  - pip install -q coverage coveralls --use-wheel
+script: py.test --cov=puppetboard --pep8  -v
+
+after_success:
+  - coveralls
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,34 @@ Changelog
 
 This is the changelog for Puppetboard.
 
+0.2.0
+=====
+
+* Full support for PuppetDB 4.x
+* Updating Semantic UI to 2.1.8
+* Updating Flask-WTF requirements to 0.12
+* Updating WTForms to 2.x
+* Restored CSRF protection on the Query Tab form
+* Updating Pypuppetdb requirement to 0.3.x
+* New configuration option OVERVIEW_FILTER allows users to add custom
+  PuppetDB query clauses to include/exclude nodes displayed on the
+  index page
+* Adding Radiator view similar to what is available in Puppet Dashboard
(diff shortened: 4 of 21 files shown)


corey-hammerton:
Have you verified that the `app.py` changes still work on a regular OS install?
Can you check the Travis CI errors?

tux-o-matic:
@corey-hammerton no breaking change with this code for a regular installation using settings.py rather than ENV variables. Same goes for static, if the location is defined in Apache or Nginx (test), that's how the resources will be served.
CI is green after indention corrections.

tux-o-matic:
A new release of PuppetBoard on PyPi will be required to use this container since it uses 'pip' internally.

corey-hammerton:
With the new `settings.py` file in this PR it appears that it contains settings specific to your environment, most notably the `PUPPETDB_HOST` setting. Because of this I cannot accept this PR. 

My suggestion is to update `README.rst` with instructions of how users can configure their containers that fit their environments.

tux-o-matic:
@corey-hammerton  I removed the local settings file and added steps to use the new Dockerfile in the README.

mterzo:
What's the purpose of this change?  The Dockerfile now installs the latest released version of puppetboard, which will only be determined at build time,  when you already have a specific version checked out. 

Also, all the performance numbers I've seen is uwsgi > gunicorn. 

I use this existing docker container for my deployment. I set the settings.py env variable which points to a volume that has my settings and keys that are needed to establish a connection.

tux-o-matic:
@mterzo this new container can do everything out of the box and can run anywhere once the image is built. The current one requires a build for every change in the settings and requires a reverse proxy to run in front of it.

As for performance, I kept this one pretty simple but for most usage I assume this gunicorn with sync worker will be enough. We could make Gunicorn use eventlet/gevent/gthread if you have scenarios with a lot of users, to which I would recommend adding a container with Nginx in front to add some caching.

mterzo:
@tux-o-matic I have multiple web services on one host the only option is a proxy. 

I want to run puppetboard:0.2.0. When I build that container I better get 0.2.0. This will get whatever the latest in pypi.

tux-o-matic:
@mterzo this new container doesn't stop you from using a reverse proxy in front of it. The choice is up to you but at least it can provide a working PuppetBoard without re-building for every host or serving static resources from a secondary service.

As for the version, VoxPupuli should publish a Docker image on the DockerHub along with every release.
So real world usage would be very simple from a usage point of view
```shell
docker run -e PUPPETBOARD_PUPPETDB_HOST=some.ip.or.name voxpupuli/puppetboard:0.2.1
```

mterzo:
@tux-o-matic gunicorn 19.6.0, doesn't serve files under /static.  I ended up having to run supervisor and nginx to host the static files.  I don't see how the static js / css files will get served.

tux-o-matic:
@mterzo if your container implementation resorts to using supervisord then something is really wrong.
My PR with a Flask route for static did work without the need for Nginx/Apache. So again maybe we should leave one PR for the ENV variables and one for a new Docker support.

mterzo:
@tux-o-matic it was another project I was trying it on. I couldn't get gunicorn to serve files out of /static even though it had it in being routed correctly in flask.  I'm concerned that newer gunicorn disabled this and now your container will have the same problem since you get the latest at build.

tux-o-matic:
As long as there is a Flask route serving content via send_from_directory(), it should work, regardless of the WSGI server in front.

mterzo:
That's what I thought. But looks like significant changes in the last few releases of gunicorn.

mterzo:
It's was flask and gunicorn hand off that I was having a problem with on another app. When using /static.  The static_folder needs to be set when it changes from the relative path static.  The app path /static is ignored running flask inside gunicorn.

mterzo:
Closing in favor of #310. Thanks @tux-o-matic you did heavy lift.

PR #280, opened 2016-07-22
Make node row in inventory link to node page

tux-o-matic:
To improve UX, made each element in an inventory row clickable to get to the node page.

PR #289, opened 2016-08-10
puppetboard/app.py: Simplifying the Inventory Code

### LINKED ISSUE #275: Simplify Inventory Code
From its initial merge the code that generates the data for the Inventory page iterates over the PuppetDB results twice. The only reason I can see that this is the case is to catch facts in nodes. 

One possible solution is to iterate over the PuppetDB facts once and generate a `dict` then in the template we can possible use [dict.get()](https://docs.python.org/2/library/stdtypes.html#dict.get) to display their values and display "undef" on KeyError

corey-hammerton:
This resolves #275

This update eliminates one iteration over the resulting inventory facts
that generates a multidimensional dictionary keyed by the node's certname
to another dictionary of key-value pairs of the fact name and fact value.

### DIFF
--- a/puppetboard/app.py
+++ b/puppetboard/app.py
@@ -291,14 +291,11 @@ def inventory(env):
     envs = environments()
     check_env(env, envs)
 
-    fact_desc = []     # a list of fact descriptions to go
+    headers = []        # a list of fact descriptions to go
     # in the table header
     fact_names = []     # a list of inventory fact names
-    factvalues = {}     # values of the facts for all the nodes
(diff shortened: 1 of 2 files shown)


tux-o-matic:
@corey-hammerton The template for the inventory page still doesn't link to the node page as presented  #280. Could you consider this improvemt fro your PR?

corey-hammerton:
@tux-o-matic I've updated this PR, does this fit with your request?

tux-o-matic:
@corey-hammerton thanks, that should do it.

PR #310, opened 2016-10-25
Default settings use environment variables

mterzo:
Update default_settings.py to behave as it exists today with it's settings, but allow for configuration to be overwritten using environment.  This allows for simpler docker deployments.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,3 +1,12 @@
-FROM grahamdumpleton/mod-wsgi-docker:python-2.7-onbuild
-CMD [ "wsgi.py" ]
+FROM python:2.7-alpine
 
+ENV PUPPETBOARD_PORT 80
+ENV PUPPETBOARD_SETTINGS docker_settings.py
+RUN mkdir -p /usr/src/app
+WORKDIR /usr/src/app
+
+COPY requirements-docker.txt /usr/src/app/
+RUN pip install --no-cache-dir -r requirements-docker.txt
+COPY . /usr/src/app
+
+CMD gunicorn -b 0.0.0.0:${PUPPETBOARD_PORT} --access-logfile=/dev/stdout puppetboard.app:app
--- a/README.rst
+++ b/README.rst
@@ -128,6 +128,20 @@ image is planned for the 0.2.x series.
 
 .. _Dockerfile: https://github.com/voxpupuli/puppetboard/blob/master/Dockerfile
 
+Usage:
+.. code-block:: bash
+  $ docker build -t puppetboard .
+  $ docker run -it -p 9080:80 -v /etc/puppetlabs/puppet/ssl:/etc/puppetlabs/puppet/ssl \
+    -e PUPPETDB_HOST=<hostname> \
+    -e PUPPETDB_PORT=8081 \
+    -e PUPPETDB_SSL_VERIFY=/etc/puppetlabs/puppetdb/ssl/ca.pem \
+    -e PUPPETDB_KEY=/etc/puppetlabs/puppetdb/ssl/private.pem \
+    -e PUPPETDB_CERT=/etc/puppetlabs/puppetdb/ssl/public.pem \
+    -e INVENTORY_FACTS='Hostname,fqdn, IP Address,ipaddress' \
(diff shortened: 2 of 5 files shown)


mterzo:
I'm running into weird issue on this.  It works great when I use these new settings and point it to the remote server.  The idea is to run the puppetboard container on the puppetdb host.  When I do I get a weird IOError 13 Permission Denied.  I run in the same container with the same settings using the dev.py it works perfectly.  It look to be something with running under WSGI.  Any thoughts?

tux-o-matic:
Couldn't we get #288 merged?
And if users may need to set extra parameters via environment, we could build on top of it.

mterzo:
TODO: I need to squash my commits.

I completely forgot about #288.

Yeah, similar concept.  If the Dockerfile was a separate repo, I could see using pip to install puppetboard.  Since the the docker file is with the code, install the checked out code into the app location.  I want to make sure we "pin" the version of gunicorn as well. 

I reordering the layering of your docker file as well so that the "common" areas a higher.

Maybe the version of gunicorn is different if found it needs to be: --access-logfile=/dev/stdout , the PR had --access-log=/dev/stdout

I don't understand why you changed the static content.  Everything worked as expected.

tux-o-matic:
@mterzo  Why duplicate so much from #288? If there had to be two PR, I'd rather have yours merged first to handle the ENV variable and then #288 for the Docker part.

Indeed Gunicorn log parameter changed between version. And yes the Docker Hub image should not just be :latest but distributed per puppetboard/gunicorn version.

I added a static route in Flask since the other routes in the app weren't meant to serve static files and expect Nginx/Apache to handle that route. In a container setup that's not the most user friendly approach.

mterzo:
I looked at cherry picking but you had several different commits. I didn't want to pull them and then squash.

Personally, I'm strict on pinning the versions of modules.  Not providing the version of gunicorn or puppetboard it's difficult to know what was packaged with the container.

With some integration with docker hub and triggers we can easily use tagging and branching to build containerized versions of of puppetboard on every commit to master without having to use pip to install puppetboard.

tux-o-matic:
GIT master always has the latest code, that's a safe assumption.
Option A:
And every time a new release is tagged on GIT then published on PyPI, the maintainer published a new image on DockerHub.
Option B:
The tagged Dockerfile pulls from GIT a certain git tag of PuppetBoard, but still uses pip for Gunicorn.

@mterzo: Do you know any mechanism to have either approaches done automatically?

mterzo:
I use B on my repos. Having A you get a slew of packages.   We need the vox account linked on docker hub.

mterzo:
@tux-o-matic I verified, It has everything to do with where the static_folder is being served.  when gunicorn is front ending flask.  flask doesn't use the app.route('/static') it uses the defines arguments passed in for static_folder and static_url.  static_url is default of '/static'.  if you want to place static content outside the app module directory you have to specify the static_folder=/path_to_content

I'm pretty happy with the way this is working.

```
docker run -it -v /my_ssl_certs:/certs \
  -e PUPPETDB_SSL_VERIFY=/certs/ca.pem -e PUPPETDB_KEY=/certs/private.pem \
  -e PUPPETDB_CERT=/certs/public.pem -e PUPPETDB_HOST=my_cool_host \
  -e PUPPETDB_PORT=8081 -p 9080:80 \
  -e INVENTORY_FACTS='Hostname,fqdn, IP Address,ipaddress' \
  -e GRAPH_FACTS="my_custom_fact" -e PUPPETBOARD_PORT=80 \
  -e "ENABLE_CATALOG=true"  puppetboard
```

mterzo:
@tux-o-matic I've been running a container off this pull request now with only gunicorn. Works great!

mterzo:
@tux-o-matic You want to take a look at this PR.  I think this solves what you were doing with #288.

mterzo:
@tux-o-matic works like a charm.

juniorsysadmin:
@mterzo @tux-o-matic Ready for merge?

mterzo:
Yep, was waiting on the hub.docker changes for #312

PR #316, opened 2016-11-18
Locally Serve Semantic-UI Required Fonts

james-powis:
This change localizes (statically stages) the google fonts included in Semetric UI which was breaking OFFLINE_MODE when there is a firewall present with drop rules.

All work is completed pending review.


james-powis:
If someone has a spare cycle to assist with getting this test written I would appreciate it, I am not really a python guy... Basically just need to test that puppetboard/static/Semantic-UI-2.1.8/semantic.min.css !~ /http/

This will help us ensure that future updating of semetric ui does not cause regression

mterzo:
Your test isn't passing. The best practice with semantic ui is to download the font from google and add it locally and serve it from there. Which isn't in the change set.

The UI is served locally in both on and offline mode. Probably should clean this up as well, so we are true to the word offline mode.

james-powis:
Tests have been completed and are functional, work complete, ready for review.

mterzo:
Sorry for the delay on this.  So I'm looking at your code, I thought this was to make offline mode actually reference local static content.

Looks like now you are just making it reference https off of a CDN instead of http.

Only good way to manage an offline mode would be to host the css locally and have the TTFs stored as static content modifying the CSS to point to the local.  I'm not sure if it's worth managing both Offline and CDN references to the content.  Offline mode should just be the default for these CSS files.

james-powis:
You are correct, and I will get that addressed shortly

james-powis:
@mterzo I have locally staged the fonts, and given i missed the HTTP reference to the font files, I have added tests. This will allow us to in the future update semantic and have tests which poke you in the eye to outline WHAT needs to be done to not break _blackhole_ installations.

mterzo:
@james-powis I'm reworking the testing code.  This will give us better validation.  This will start using the FlaskTest client which will give us the ability to read what's served instead of what's on disk.  This will allow for us to test offline mode vs online mode.

nikdoof:
Am I correct saying this has now been resolved by #397?

mterzo:
Yup.

mterzo:
Needed the CSS for cloud not to change. So needed some python flask magic to serve static files.  I was looking for this test code thought it was already merged.

PR #318, opened 2016-12-02
radiator column width to percent

ts-mini:
When using more than 1000 nodes>
clipping
The count values get clipped on all rows

But setting to 100% width produces the following:
100_percent


mterzo:
What browser/OS are you viewing. The fonts don't look correct.

ts-mini:
Chrome Version 54.0.2840.100 (64-bit) 

Distributor ID:	Ubuntu
Description:	Ubuntu 16.04.1 LTS
Release:	16.04
Codename:	xenial

mterzo:
Will need to look at this on other platforms. Thanks for the pr.

Issue #303, opened 2016-09-20
JSON of Radiator view


peelman:
the Radiator view is a great addition to v0.2.0, thank you for that.  One tweak I'd ask is a JSON endpoint (basically respond to server.tld/radiator.json) that returns the same data, just unformatted and in raw JSON.

Any thoughts?

mterzo:
#319 added this feature.

peelman:
This doesn't seem to have made it into the 0.2.1 release.

mterzo:
it was. Set the accept header to application/Json and it will return Json data.

PR #319, opened 2016-12-04
Add JSON endpoint to Radiator view

llabrat:
Add /radiator.json and /env/radiator.json to provide json output from
Radiator view as requested in #303

### DIFF
--- a/puppetboard/app.py
+++ b/puppetboard/app.py
@@ -1031,7 +1031,9 @@ def catalog_compare(env, compare, against):
 
 
 @app.route('/radiator', defaults={'env': app.config['DEFAULT_ENVIRONMENT']})
-@app.route('/radiator.json', defaults={'env': app.config['DEFAULT_ENVIRONMENT'], 'to_json': True})
+@app.route('/radiator.json',
+           defaults={'env': app.config['DEFAULT_ENVIRONMENT'],
+                     'to_json': True})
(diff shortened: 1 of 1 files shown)


puppetboard/app.py
```diff
@@ -1031,8 +1031,13 @@ def catalog_compare(env, compare, against):
...
 
 @app.route('/radiator', defaults={'env': app.config['DEFAULT_ENVIRONMENT']})
+@app.route('/radiator.json', endpoint='radiator_json',
+           defaults={'env': app.config['DEFAULT_ENVIRONMENT'],
+                     'to_json': True})
+@app.route('/<env>/radiator.json', endpoint='env_radiator_json',
+           defaults={'to_json': True})
 @app.route('/<env>/radiator')
-def radiator(env):
+def radiator(env, to_json=False):
```
mterzo: Instead of adding a to_json lets use the Accept header:

    if ('Accept' in request.headers
        and request.headers["Accept"] == 'application/json'):
        return jsonify(**stats)

mterzo: @llabrat Any comment on this suggestion?  I think this is a cleaner approach.

llabrat: Sorry, been on vacation and still catching up. I agree, I will update.

mterzo: No worries.

PR #320, opened 2016-12-07
[Security] Implement bandit

roidelapluie:
Bandit (https://github.com/openstack/bandit) is a python
security linter.

Signed-off-by: Julien Pivotto <roidelapluie@inuits.eu>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,9 @@ install:
   - pip install -r requirements.txt
   - pip install -r requirements-test.txt
   - pip install -q coverage coveralls --use-wheel
-script: py.test --cov=puppetboard --pep8  -v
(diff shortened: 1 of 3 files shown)


petems:
This seems simple, safe and a benefit for catching any AppSec issues. LGTM 👍

mterzo:
Safe simple, shell script in necessary. Also is there a way to collect results?

roidelapluie:
what do you mean collecting results? if we introduce security problems it will fail the build

mterzo:
So completely binary pass/fail. No this is a deprecated warning. As such?

roidelapluie:
Indeed. If we have false positive we can use control comments.

PR #322, opened 2016-12-14
Introducing simple plugin system for Puppetboard (rebase)

petems:
This is a rebase of https://github.com/voxpupuli/puppetboard/pull/168

Haven't had a response from @marcinzaremba since he opened back in Oct 2015, but this seems like a great feature.

As @nibalizer mentioned last PR, would be nice to have an example plugin. I'm not a python dev but with a bit of a assistance I could try implementing something basic?


mterzo:
@petems Thanks for the rebase, unfortunately you need to do it again default_settings was changed.  I agree with @nibalizer having a sample plugin would be nice.  Of course @daenney also mentioned about a bit of a refactor to work.  I need to go work with flask blueprints more before I can assist with this.

daenney:
This is a bit of a hacky-whacky way to do plugins. It works but leveraging setuptools or even better, [PluginBase](http://pluginbase.pocoo.org) would be nicer. I don't like the manual import stuff there, that's prone to errors. 

Also, the overriding of `app.config['PLUGINS']` once we've loaded it from config to a list of initialised blueprints is bothersome. We shouldn't change the config representation at runtime that way.

petems:
Closing out old PR's, this is many years after the fact so I'm assuming it's super out of date! Closing 👍

PR #325, opened 2016-12-29
catch a division by zero in radiator while environment has no nodes

rmdir:
In radiator, when an environment as no nodes, num_nodes is null. This patch catch the divisions by zero and set the results to 0 in this case

### DIFF
--- a/puppetboard/app.py
+++ b/puppetboard/app.py
@@ -1097,20 +1097,22 @@ def radiator(env):
             stats['unchanged'] += 1
 
     try:
-        stats['changed_percent'] = int(100 * stats['changed'] / float(num_nodes))
+        stats['changed_percent'] = int(100 * (stats['changed'] /
+                                              float(num_nodes)))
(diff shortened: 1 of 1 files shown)


mterzo:
I get your problem.  If nodes are 0, which is causing the division by zero error. shouldn't the other values be 0 as well?

so,  if number of nodes == 0 then set nodes to 1, and it all should show up as 0% making the code a bit simpiler.

mterzo:
So I did a quick test, my assumption was correct, but the problem is that number of nodes reported is 1, instead of 0.

The only way I was able to reproduce this was to create a node in an environment, then delete that 1 node from that environment.  Is that the scenario you have?

mterzo:
I would like to pull this in, can you squash this into 1 commit?

corey-hammerton:
@mterzo Github has a "Squash and merge" option, along with "Create a merge commit" and "Rebase and merge"

mterzo:
@corey-hammerton thank you, i never noticed.

rmdir:
@mterzo Thanks for the commit.

Issue #304, opened 2016-09-20
Offline mode still pulls content from web


abrownfi:
Semantic UI still relies on web-resources, regardless of whether offline mode is set.

The pull is initiated in semantic.min.css:  @import url(https://fonts.googleapis.com/css?family=Lato:400,700,400italic,700italic&subset=latin);

That URL results in CSS that further pulls down the individual woff files.

corey-hammerton:
Semantic UI is a third-party framework that we simply use. We can make changes to our code for CSS Importing of font files but the contents of the Semantic UI files are maintained by a third-party organization. As a result we will not be making any modifications to any of these files.

Closing this issue with a wont-fix label.

bastelfreak:
Hi,
in my opinion the parameter offline_mode is a bit misleading. As a user I would expect that it completely works while having no internet connection, not only partly. @james-powies suggested on IRC that he can provide a patch. Should we give it a try?

james-powis:
Being offline is all find and well, and being behind a firewall with outbound reject policies works well too. 

However a firewall policy where the action is to drop requires http to timeout before each page of puppetboard will properly load. This leads to the argument "change your firewall policy to reject", doing so actually causes more problems because `netstat -an | grep SYN_SENT` helps identify issues such as this, and will not work with reject.

james-powis:
basic monkey patch is done, just pending getting successful tests written to prevent regression:
https://github.com/voxpupuli/puppetboard/pull/316

juniorsysadmin:
Changed label to bug since 'offline' is not as one would expect. Patching third-party frameworks is definitely a nono though.

mterzo:
@juniorsysadmin Yeah, patching 3rd party sucks.  Here's the suggestion: 
https://github.com/Semantic-Org/Semantic-UI/issues/1521

PR #326, opened 2017-01-03
Flask testing.

mterzo:
Working on better testing with flask.  Mocking out puppetdb some so we the initial page can load.

Updating some templates to have ending tags for better parsing.


mterzo:
Added tests for #304.

I'm not looking at the CSS files that are being serviced, but look at the CSS and JS links that are provided in the html that's being served back.

mterzo:
I added testing for #325.  Right now it's allowed to fail since the change that's the PR fixes the division by zero error.

PR #327, opened 2017-01-05
Use puppetdb as PUPPETDB_HOST on Docker

raphink:
This allows to plug and play a PuppetDB container (or use extra host)


mterzo:
Thanks. Yeah it will never be localhost.

PR #328, opened 2017-01-08
Setup.py : zip_safe to permit static serve

redref:
During setup (apache + wsgi), I got errors about templates ("File not found").

Found this parameter to solve it. It prevents pip/easy_install to zip python module (used by egg installation mode) in installation directory.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -28,6 +28,7 @@ setup(
     description='Web frontend for PuppetDB',
     include_package_data=True,
     long_description='\n'.join((README, CHANGELOG)),
+    zip_safe=False,
     install_requires=[
         "Flask >= 0.10.1",
         "Flask-WTF >= 0.12, <= 0.13",


mterzo:
I haven't run into this before so I had to do some reading.

http://peak.telecommunity.com/DevCenter/setuptools#setting-the-zip-safe-flag

If your building from the git tree you can build with `setup.py easy_install --zip-ok` to produce your zip file that will support the static serve without the modification for the source code.

@corey-hammerton, we never talked about if you were doing anything special to publish the python module.  I typically only release egg and whl.

redref:
I did not managed to catch clearly your opinion on this. But, to be clear, my problem came the other way. I would have needed `setup.py easy_install --always-unzip` to get it working.

Statics must certainly be served from an unzipped directory (not sure on templates). I think according to your link, this flag should be explicit.

mterzo:
I'm just trying to understand how we have released packages in the past.  I think this should be a good update.  Trying to understand why it works from pypi without the change.

PR #323, opened 2016-12-19
Paging - Revamp tables with Jquery Datatables (Ajax)

redref:
Hi,

I worked on paging and hopes it can improve actual paging features. Not sure about it, but different anyway.

Right now, there is ugly things (inline JS, ...) and this may be extended to all pages. But, please give it a try and let me know your thoughts and/or any improvements you may find suitable.

PR #341, opened 2017-01-30
Update HTML and Javascript

### LINKED ISSUE #159: Donut view for node states
It would be great to be able to visualize node states with a donut view (like Foreman).

This can easily be achieved using c3js (already used in Puppet board): http://c3js.org/samples/chart_donut.html

I would assign this to myself, just as a reminder to code it someday (unless someone else wants to do it).

### LINKED ISSUE #340: Breadcrumb UI overlap on node.
Paging for reports now overlaps one node.

breadcrumb_too_wide
breadcrumb_overlap

mterzo:
Fix #340, by changing the pagination style of the report on the node.
Fix #159 by adding new configuration setting allowing the graph type to be changed.

After the PR of #323 all the old non-data tables were no longer sorting and filtering.  This is fixed.


redref:
Hi @mterzo ,
I'm fine with simple paging and disable search on node page. Previous behavior was to show 10 reports and the "Show All" to reports page.

On the other end, I'll check fact sorting and filtering. In my mind, I'm pretty sure I checked tables.js was not used anymore, but I certainly missed it. I will prefer to extend datatables use right away. Will open a PR to your branch on this subject.

mterzo:
Thanks for your help on this.

mterzo:
@redref I think this is ready to go, any adjustments?

Issue #333, opened 2017-01-26
Display the number of resources for each node


thias:
I'm migrating Puppet Masters, from Puppet 3 to Puppet 4, and also migrating from Puppet Dashboard to Puppetboard. I've looked everywhere, and Puppetboard doesn't seem to show the number of resources per node report. It has the total resources and a global average per node on its index, but nothing more after that.

In Puppet Dashboard, each run's report showed the Total number of resources, then Pending, Changed and Unchanged. Very useful. Even more useful, as I was planning on comparing the number of resources for each node on the old and new masters, as I have hit "silent" issues during past migrations, where some code was no longer properly applied, resulting in no changes, but resulting in less resources being managed.

In Puppetboard, next to each node I see 3 numbers, but they're always "0", except the middle one which does count the "Changed" when that happens. Looking at the README's screenshots, it seems like the left one is to count the failed resources. I can't figure out what the right one is.

I would really like to see the total number of resources for each run in the Overview, Nodes or Reports pages... am I the only one?

mterzo:
Any chance you could show a screenshot?

thias:
Sure, here's what the details of my nodes with pending changes looks like :
![puppetdashboard](https://cloud.githubusercontent.com/assets/787766/22340907/4b5b3496-e3ef-11e6-82b1-93bd938a8be9.png)

mterzo:
Thanks, that doesn't look too bad.  The data is there, just need to display it.

thias:
FYI, this is the similar output from Puppetboard... the 0 0 0 isn't very useful... for the PENDING nodes, it would be really useful for one of these values to be the number of resources pending changes.
![puppetboard](https://cloud.githubusercontent.com/assets/787766/22342336/54e03962-e3f4-11e6-90ea-d05cd0fbda36.png)

redref:
Hi,

Do you think this can be covered by #344 ? Right now, we are talking about 'noop' (aka pending).

Do you think total is mandatory ?

mterzo:
Yeah adding a pending would be useful.

thias:
BTW, poking around and now that I have ~200 nodes reporting, I have figured out what the 3 numbers mean : 
* Left : Failed
* Middle : Changed
* Right : Skipped (so only present when some "Failed" are too, AFAICS)

I have added a comment to #344, as having the details in Reports would be nice, but also in Overview would be even better.

mterzo:
[...]

PR #344, opened 2017-01-31
Add total and noop boxes on reports page.

redref:
Sum up 'noop' and 'skipped' resources from reports into displayed "skips" value. Particularly relevant when doing 'noop' runs on nodes in order to identify how far the node is from baseline.

'noop' resources may also be explicitly declared in catalogs (even in normal run). In this case, we are losing a bit of information. But, I don't know what is the typical use case of noop resource. This use case may match our interface behavior.

### DIFF
--- a/puppetboard/static/css/puppetboard.css
+++ b/puppetboard/static/css/puppetboard.css
@@ -72,6 +72,14 @@ h1.ui.header.no-margin-bottom {
   background-color: #DB843D;
 }
 
+.ui.header.total {
+  color: #989898;
+}
+
+.ui.label.total {
+  background-color: #989898;
+}
+
 .ui.label.unchanged {
   background-color: #89A54E;
 }
--- a/puppetboard/static/css/puppetboard.css
+++ b/puppetboard/static/css/puppetboard.css
@@ -44,7 +44,7 @@ h1.ui.header.no-margin-bottom {
   color: #AA4643;
 }
 
-.ui.label.failed {
+.ui.label.failed, .ui.label.events.failure {
   background-color: #AA4643;
 }
 
@@ -52,7 +52,7 @@ h1.ui.header.no-margin-bottom {
   color: #4572A7;
 }
 
-.ui.label.changed {
+.ui.label.changed, .ui.label.events.success {
   background-color: #4572A7;
 }
 
@@ -68,15 +68,11 @@ h1.ui.header.no-margin-bottom {
   color: #DB843D;
 }
 
-.ui.label.noop {
+.ui.label.noop, .ui.label.events.noop {
   background-color: #DB843D;
 }
 
-.ui.header.total {
-  color: #989898;
-}
-
-.ui.label.total {
+.ui.label.resources.total {
   background-color: #989898;
 }
 
@@ -88,7 +84,7 @@ h1.ui.header.no-margin-bottom {
   color: orange;
 }
 
-.ui.label.skipped {
+.ui.label.skipped, .ui.label.resources.skipped {
   background-color: orange;
 }
 
--- a/puppetboard/app.py
+++ b/puppetboard/app.py
@@ -537,7 +537,7 @@ def reports_ajax(env, node_name):
         if total is None:
             total = puppetdb.total
 
-        metrics[report.hash_] = {'resources': {}, 'events': {}}
+        metrics[report.hash_] = {}
         for m in report.metrics:
             if m['category'] not in metrics[report.hash_]:
                 metrics[report.hash_][m['category']] = {}
--- a/puppetboard/app.py
+++ b/puppetboard/app.py
@@ -531,25 +531,17 @@ def reports_ajax(env, node_name):
         reports_events = []
         total = 0
 
-    report_event_counts = {}
-    # Create a map from the metrics data to what the templates
-    # use to express the data.
-    events_metrics = ['success', 'failure', 'noop']
-    resources_metrics = ['total', 'skipped']
+    # Convert metrics to relational dict
+    metrics = {}
     for report in reports_events:
         if total is None:
             total = puppetdb.total
 
-        report_counts = {}
-        for metric in report.metrics:
-            if ('category' in metric and metric['category'] == 'events' and
-               'name' in metric and metric['name'] in events_metrics):
-                report_counts[metric['name']] = metric['value']
-            if ('category' in metric and metric['category'] == 'resources' and
-               'name' in metric and metric['name'] in resources_metrics):
-                report_counts[metric['name']] = metric['value']
-
-        report_event_counts[report.hash_] = report_counts
+        metrics[report.hash_] = {'resources': {}, 'events': {}}
+        for m in report.metrics:
+            if m['category'] not in metrics[report.hash_]:
+                metrics[report.hash_][m['category']] = {}
+            metrics[report.hash_][m['category']][m['name']] = m['value']
 
     if total is None:
         total = 0
@@ -560,7 +552,7 @@ def reports_ajax(env, node_name):
         total=total,
         total_filtered=total,
         reports=reports,
-        report_event_counts=report_event_counts,
+        metrics=metrics,
         envs=envs,
         current_env=env,
         columns=REPORTS_COLUMNS[:max_col])
--- a/puppetboard/app.py
+++ b/puppetboard/app.py
@@ -534,21 +534,20 @@ def reports_ajax(env, node_name):
     report_event_counts = {}
     # Create a map from the metrics data to what the templates
     # use to express the data.
-    report_map = {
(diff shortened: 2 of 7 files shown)


redref:
@mterzo What do you think about this ?

mterzo:
I think it should be it's own thing. Occurs not always in a noop situtiaton. When a resource is skipped based on a schedule criteria then skipped is incremented.

redref:
Ok. If we are fine to add a forth (grey?) box, I will modify this PR.

mterzo:
Actually I would like to figure out how to show total resources as well as part of #333. So adding to the list of boxes would be required.

redref:
Working on it. I'm searching a good color.

mterzo:
I'm concerned that having 5 boxes there would make it some what confusing with out a column header or maybe a mouse over?  Almost wonder if the Td of status and counts should be combined and label the the number like total: x , changed:x. And so on.

redref:
Agree to split up TDs. Will also permit ordering on it.

redref:
Digging into, ordering on metrics / events is not possible. Roll-backed to "on hover" field description. Still, it is possible to add columns and disable sorting on it.

Pushed a version working on reports page. 

Values are not available from nodes endpoint. Right now, it's better to let nodes/overview like they are. I plan to convert those pages to paging using 'last_report?' from nodes_endpoint.

mterzo:
Looking at the change, I was gonna create a dictionary from the array after receiving the events
turning:

```
[{
   'type': 'resource',
  'name': 'somelabel'
  'value': 'x'
}, 
...]
```
to:
```
{ 
  'resources': {
     'somelabel': value',
     'other_resource': 'value',
  },
 'events': {
   'total': 'value',
  'noop': 'value',
  }
}
```

Which makes it easy for
```
total_resource = mapped_data.get('resources', {}).get('total', 0)
noops = mapped_data.get('events', {}.get('noop', 0)
```

thias:
This is very interesting! I would nevertheless still really like being able to see the number of pending resources changes for NOOP reports on the Overview main page, not just the Reports page.

Quick explanation : The Radiator is nice to have on a large screen, and the Reports are nice to go and dig for details, but the Overview is what our sysadmins have open all day, proactively looking at any/all deviations reported by noop runs (in addition to failed, changed and unreported, which are all equally important). It's extremely useful to know the scope of these pending changes. For instance a mass deploy of a minor change might immediately create 2 pending changes on all nodes, while others more specific might create more pending changes but only on a few nodes : For me, it's extremely useful to be able to spot the nodes that have more than 2 in the Overview, then go to the detailed reports from there.

If the number of counters next to each node in the Overview should be kept to a minimum by default, then maybe it could be configurable? Different ways of using Puppet justify having different importance for these values. For instance, I don't really care about the "skipped", as those happen only for "failed" anyways. OTOH, I would really like to see the total number of resources as well as the number of pending changes, something that people not using noop runs certainly don't care about.

thias:
@mterzo : I really like your suggestion. It would be a lot simpler to keep the structure and naming as close to what pypuppetdb provides, as it's currently quite confusing to be accessing 'skips' in one place, 'skipped' in another, etc...

Playing around with the code, it was fairly easy to get `events['noops']` to appear where skipped was, but I still can't seem to figure out how to get a 'total' (or even if it's actually there, or even if noops+successes+failures+skips should be equal to the total of managed resources... as it seems not).

As an "end user" of Puppetboard, I would find it pretty neat to be able to configure something like this : 
```
OVERVIEW_NODE_DETAIL_COUNTERS = [ 'total', 'noops', 'successes', 'failures' ]
```
...just sayin' :-)

redref:
Just to make things clear. Here are the metrics (Report type) we get 

`
ERROR:puppetboard.app:{u'category': u'resources', u'name': u'changed', u'value': 0}
ERROR:puppetboard.app:{u'category': u'resources', u'name': u'corrective_change', u'value': 0}
ERROR:puppetboard.app:{u'category': u'resources', u'name': u'failed', u'value': 1}
ERROR:puppetboard.app:{u'category': u'resources', u'name': u'failed_to_restart', u'value': 0}
ERROR:puppetboard.app:{u'category': u'resources', u'name': u'out_of_sync', u'value': 1}
ERROR:puppetboard.app:{u'category': u'resources', u'name': u'restarted', u'value': 0}
ERROR:puppetboard.app:{u'category': u'resources', u'name': u'scheduled', u'value': 0}
ERROR:puppetboard.app:{u'category': u'resources', u'name': u'skipped', u'value': 0}
ERROR:puppetboard.app:{u'category': u'resources', u'name': u'total', u'value': 9}
ERROR:puppetboard.app:{u'category': u'time', u'name': u'config_retrieval', u'value': 1.642247005}
ERROR:puppetboard.app:{u'category': u'time', u'name': u'exec', u'value': 0.000898299}
ERROR:puppetboard.app:{u'category': u'time', u'name': u'filebucket', u'value': 9.2661e-05}
ERROR:puppetboard.app:{u'category': u'time', u'name': u'schedule', u'value': 0.00037461}
ERROR:puppetboard.app:{u'category': u'time', u'name': u'total', u'value': 1.643612575}
ERROR:puppetboard.app:{u'category': u'changes', u'name': u'total', u'value': 0}
ERROR:puppetboard.app:{u'category': u'events', u'name': u'failure', u'value': 1}
ERROR:puppetboard.app:{u'category': u'events', u'name': u'success', u'value': 0}
ERROR:puppetboard.app:{u'category': u'events', u'name': u'noop', u'value': 1}
ERROR:puppetboard.app:{u'category': u'events', u'name': u'total', u'value': 1}
`

From what we discussed, we need to select resources.total, events.failure, events.success, resources.skipped, events.noop.

I totally agree with you, we need those numbers on nodes and overview pages. But right now, it would be really difficult to bring them in (with actual code base). I'm pushing this change just like an example (template) of what we want.

Then, we can work on overview/nodes pages revamp which is mandatory for me. Right now, I cannot use in production cause overview page is longer than 10sec to load.

thias:
@redref : Thanks for the details, I see where you're going with this, and it makes sense (at least to me). I'm guessing that your idea is then to migrate all of the other "events" to use this new "metrics".

So here's my updated wishlist item :-)
```
OVERVIEW_NODE_DETAIL_COUNTERS = [ 'resources.total', 'events.noops', 'events.successes', 'events.failures' ]
```
...bonus points for making it possible to set 'time.total' there, to get useful feedback about puppet run duration!

mterzo:
@thias, there's a line between having enough or too much information on the dashboard.

redref:
Screen capture with time.total activated.

mterzo:
That's busy.

redref:
I'm in vacations :)

mterzo:
Yeah, you too.  There's a lot going in that table now.

redref:
Rebased.

Just noticed, your last test change (aka status_report) broke some tests.

redref:
Updated pypuppetdb on my side. Tests are working.

thias:
@redref : Very nice! For my nodes, the output of the resources.total shows as "4.6e+02", "3.1e+02", etc. though. Given your screenshot, I'm guessing it didn't apply to you because of your single digit values.

redref:
@thias this one should do the trick !

mterzo:
The browser needs to start getting really wide with all these extra columns.

mterzo:
I would like to do a release of 0.2.1 tomorrow 2/3.  I'm not sure this is quite ready yet.  There are a bunch of features and bug fixes already there, I think we can start releasing quicker instead of every 7 months.  Do you have any objections to that?

redref:
I think as we have now a configuration option, this should do fine.

With or without this one, it's fine to release. To cover my needs and thias need, I will need also a pypuppetdb release (nodes change). Better to wait next train.

mterzo:
I think you may have to:

```
git checkout master
git fetch upstream
git merge upstream/master
git push master
git checkout <branch>
git pull --rebase origin master
```

mterzo:
That help?  the other PR has a similar problem.

redref:
Reworked the tree !

Next time, I will do it without the fancy github green buttons !

puppetboard/static/css/puppetboard.css
```diff
@@ -44,15 +44,15 @@ h1.ui.header.no-margin-bottom {
   color: #AA4643;
 }
 
-.ui.label.failed {
+.ui.label.failed, .ui.label.events.failure {
```
mterzo: What happens on labels we don't know about so something like events.foobar (if foobar was in events list).

redref: Going to grey by default.

IMO, it's fine this way. Cover relevant colors and default to additionnals.

dominiquearpin:
Any news for this PR?

Issue #346, opened 2017-02-01
Internal Server Error, TemplateSyntaxError: non-default argument follows default argument


thias:
This is with f016820, latest commit on master as of right now (continuation of #345, same environment as that report).

For the "Overview" initial page, I get the top menu, but the content of the page is an "Internal Server Error" message, with the following logged : 
```
[2017-02-01 09:48:39,613] ERROR in app: Exception on / [GET]
Traceback (most recent call last):
File "/home/puppetboard/.local/lib/python2.7/site-packages/flask/app.py", line 1982, in wsgi_app
response = self.full_dispatch_request()
File "/home/puppetboard/.local/lib/python2.7/site-packages/flask/app.py", line 1614, in full_dispatch_request
rv = self.handle_user_exception(e)
File "/home/puppetboard/.local/lib/python2.7/site-packages/flask/app.py", line 1517, in handle_user_exception
reraise(exc_type, exc_value, tb)
File "/home/puppetboard/.local/lib/python2.7/site-packages/flask/app.py", line 1612, in full_dispatch_request
rv = self.dispatch_request()
File "/home/puppetboard/.local/lib/python2.7/site-packages/flask/app.py", line 1598, in dispatch_request
return self.view_functions[rule.endpoint](**req.view_args)
File "/home/puppetboard/.local/lib/python2.7/site-packages/puppetboard/app.py", line 256, in index
current_env=env
File "/home/puppetboard/.local/lib/python2.7/site-packages/flask/templating.py", line 134, in render_template
context, ctx.app)
File "/home/puppetboard/.local/lib/python2.7/site-packages/flask/templating.py", line 116, in _render
rv = template.render(context)
File "/home/puppetboard/.local/lib/python2.7/site-packages/jinja2/environment.py", line 1008, in render
return self.environment.handle_exception(exc_info, True)
File "/home/puppetboard/.local/lib/python2.7/site-packages/jinja2/environment.py", line 780, in handle_exception
reraise(exc_type, exc_value, tb)
File "/home/puppetboard/.local/lib/python2.7/site-packages/puppetboard/templates/_macros.html", line 93, in template
{% macro datatable_init(table_html_id, ajax_url, default_length, length_selector, extra_options=None, columns) -%}
TemplateSyntaxError: non-default argument follows default argument
[...]

PR #347, opened 2017-02-01
Fix 'TemplateSyntaxError: non-default argument follows default argument'

thias:
This fixes the problem I reported in  #346, though the "Reports" page doesn't seem to work, it always just shows "Processing..." for me.


mterzo:
This is already being removed. per #341

thias:
I can confirm that this is now fixed in the master branch, thanks!

Issue #345, opened 2017-02-01
Puppetboard doesn't start because of @app.errorhandler(204) error


thias:
This is with f016820d3a18267f888555cdb046d3d2e8eb3ed8, latest commit on master as of right now.

I just ran : `pip install --user git+https://github.com/voxpupuli/puppetboard.git@master`

And got the following : 
```
[puppetboard@peanut site-packages]$ ls
click                             puppetboard
click-6.7.dist-info               puppetboard-0.2.0-py2.7.egg-info
flask                             pypuppetdb
Flask-0.12.dist-info              pypuppetdb-0.3.2.dist-info
flask_wtf                         requests
Flask_WTF-0.13-py2.7.egg-info     requests-2.13.0.dist-info
itsdangerous-0.24-py2.7.egg-info  test
itsdangerous.py                   tests
itsdangerous.pyc                  werkzeug
jinja2                            Werkzeug-0.11.15.dist-info
Jinja2-2.9.5.dist-info            wtforms
markupsafe                        WTForms-2.1-py2.7.egg-info
MarkupSafe-0.23-py2.7.egg-info
```
I've configured Puppetboard to run from a systemd service (this is RHEL7) : 
```
[Service]
User=puppetboard
WorkingDirectory=/home/puppetboard
ExecStart=/sbin/uwsgi --socket :3000 --plugin python --wsgi-file uwsgi.py
```
First issue is that Puppetboard doesn't even start : 
```
Starting Puppetboard...
*** Starting uWSGI 2.0.14 (64bit) on [Wed Feb  1 09:40:06 2017] ***
compiled with version: 4.8.5 20150623 (Red Hat 4.8.5-11) on 25 January 2017 15:08:40
os: Linux-3.10.0-514.6.1.el7.x86_64 #1 SMP Sat Dec 10 11:15:38 EST 2016
nodename: peanut.example.com
machine: x86_64
clock source: unix
pcre jit disabled
detected number of CPU cores: 8
current working directory: /home/puppetboard
detected binary path: /usr/sbin/uwsgi
*** WARNING: you are running uWSGI without its master process manager ***
your processes number limit is 31226
your memory page size is 4096 bytes
detected max file descriptor number: 1024
lock engine: pthread robust mutexes
thunder lock: disabled (you can enable it with --thunder-lock)
uwsgi socket 0 bound to TCP address :3000 fd 3
[...]

PR #348, opened 2017-02-02
Support for latest version

mterzo:
Adding "Unpinned" upgrade checks on packages that puppetboard depends on.  These unpinned checks are allowed to fail, but will identify when a new release of a package will break puppetboard.

Adding fixes for newer Flask, Jinja2 and pep8 issues.  Fixing #345

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,25 @@ python:
   - "2.7"
   - "3.5"
   - "3.6"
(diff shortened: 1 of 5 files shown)


thias:
Nice! I can confirm this fixes #345 as well as the non-working "Reports" page issue I was having.

PR #349, opened 2017-02-02
Fix dailychart in python3 + write a test for it

redref:
Everything in the title.


mterzo:
I don't understand why have 2 modes when it never gets called the other way.

redref:
Just in case, we wanted to iterate the other way. We may put this method into utils later.

redref:
Let me know.

redref:
This must be in the release IMO.

mterzo:
@redref can you figure out why there are 6 commits on this?

redref:
Much better !

PR #353, opened 2017-02-03
Revamp catalog page with paging

redref:
Revamped in a two step selection.

### DIFF
--- a/puppetboard/app.py
+++ b/puppetboard/app.py
@@ -40,6 +40,12 @@ REPORTS_COLUMNS = [
      'name': 'Agent version'},
 ]
 
+CATALOGS_COLUMNS = [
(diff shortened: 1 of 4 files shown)


mterzo:
Awesome test code!  Now to add some mocks to get data out of faked pypuppet.

redref:
Added some testing. But, I don't test paging yet. I have some ideas to extend testing on this point. If you agree, I plan to do it with my coming 'Overview / Nodes paging revamp' PR as it will require changes to 'mock_puppetdb_default_nodes'.

Removed skipped node as this status is not ever coming from pypuppetdb.

mterzo:
Yeah. Shouldn't be too bad. I was looking at phantomjs as well.  Wondering how to test all this JavaScript.

PR #355, opened 2017-02-04
Inventory page revamp with paging

redref:
Same one here about inventory page.

Unfortunately, ordering is not working on facts value (nor on /facts, nor on /fact-contents, despite last one should support it according to puppetdb documentation).

Workaround is to load the overall data in one call with "serverSide" set to off in datatables. Paging/Sorting is still working on client side.

Load times (on my test platform @6k nodes)

  * previous page (4s load, 4s DOM treatment, 1s sorting)
  * now (<500ms load, 2,5s json query, <100ms DOM with 100 lines, <100ms sorting)

### DIFF
--- a/puppetboard/app.py
+++ b/puppetboard/app.py
@@ -418,12 +418,12 @@ def inventory_ajax(env):
     total = len(fact_data)
 
     return render_template(
-            'inventory.json.tpl',
-            draw=draw,
-            total=total,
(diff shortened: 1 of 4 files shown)


redref:
This solve #361 pretty straightforward.

redref:
Rebased.

Issue #302, opened 2016-09-07
Facts not showing (blank screen)


chris-bmj:
For example the 'apt_has_updates' fact from the puppetlabs apt module.  It works at the command line, returning a boolean, but puppetboard shows a blank screen. Apache log shows the following:

[Wed Sep 07 09:09:22.373939 2016] [:error] [pid 12070:tid 140324810737408] [client 10.1.172.85:58784] UndefinedError: 'bool object' has no attribute 'replace', referer: http://puppetboard.company.com/sandpit/facts

This is not limited to booleans.  The 'apt_security_updates' fact returns an int, and gets the following error:

[Wed Sep 07 09:11:22.646618 2016] [:error] [pid 12070:tid 140324877879040] [client 10.1.172.85:58840] UndefinedError: 'int object' has no attribute 'replace', referer: http://puppetboard.company.com/sandpit/facts

Regards
Chris

chris-bmj:
I've found out that this only happens if the problem fact appears in the GRAPH_FACTS array.

juniorsysadmin:
Anyone else able to reproduce and verify as a bug?

schildwaechter:
We ran into the same issue with the `processorcount` fact where we saw an empty facts page but no error.
After dropping it from `GRAPH_FACTS` it worked again.

smortex:
Here is a full backtrace when trying to access `/fact/is_pe` with `is_pe`  included into `GRAPH_FACTS`:

```
App 8650 stderr: [ pid=8650, time=2017-02-05 15:47:44,693 ]: WSGI application raised an exception!
App 8650 stderr: Traceback (most recent call last):
App 8650 stderr:   File "/usr/local/lib/ruby/gems/2.3/gems/passenger/src/helper-scripts/wsgi-loader.py", line 163, in main_loop
App 8650 stderr:     socket_hijacked = self.process_request(env, input_stream, client)
App 8650 stderr:   File "/usr/local/lib/ruby/gems/2.3/gems/passenger/src/helper-scripts/wsgi-loader.py", line 302, in process_request
App 8650 stderr:     for data in result:
App 8650 stderr:   File "/usr/local/lib/python2.7/site-packages/werkzeug/wsgi.py", line 703, in __next__
App 8650 stderr:     return self._next()
App 8650 stderr:   File "/usr/local/lib/python2.7/site-packages/werkzeug/wrappers.py", line 81, in _iter_encoded
App 8650 stderr:     for item in iterable:
[...]

PR #356, opened 2017-02-05
Unbreak boolean facts graph

smortex:
When graphing boolean facts, a `UndefinedError: 'bool object' has no attribute 'replace'` exception is raised (#302).

This patch attempts to turn whatever is going to be displayed to a string.

### DIFF
--- a/puppetboard/templates/fact.html
+++ b/puppetboard/templates/fact.html
@@ -7,7 +7,7 @@ var chart = null;
 var data = [
 {% for fact in facts|groupby('value') %}
   {
-  label: '{{ fact.grouper.replace("\n", " ") }}',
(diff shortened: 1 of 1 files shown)


smortex:
Sorry for the late follow-up.

I don't do Python, and have no idea about how to test this with the current test infrastructure. Feel free to integrate such tests, in the meantime I will add the fix to the FreeBSD port which I maintain.

@redref seems to be doing something related to this issue.  I don't understand what happens there, but if it makes this PR irrelevant maybe we can close it?

tux-o-matic:
@mterzo I tested this simple fix and it works, pie charts can also be displayed on boolean Facts listed under GRAPH_FACTS.

PR #357, opened 2017-02-05
Facts revamp with datatables.

redref:
Like inventory, ordering by fact value is not possible for facts.

On /facts - 
 - removed environment filtering as this request was really long with a lot of nodes (now, it would be possible to get a link to an empty table)
 - reworked column repartition in python to fix jinja >= 2.9 behavior (https://github.com/pallets/jinja/issues/641)

On /fact
 - push everything into datatables (AJAX) - which save much time in DOM parsing and keep interface responsive during loading time
 - make datatables JSON call to also feed C3.js chart data if chart enabled

On /node
 - switch to datatables
 - mutualize JSON call code with /fact backend endpoint


redref:
Port of #356 fix with the new template.

Add this case to the tests.

redref:
On #356 subject.

Python3 returned me 
```
    return sorted(map(_GroupTuple, groupby(sorted(value, key=expr), expr)))
TypeError: unorderable types: bool() < str()
```

Solved by adding a jinja2 filter used in a map before groupby.

redref:
Rebased.

tux-o-matic:
JSON is a first class citizen in Python, why choose to generate JSON in a template rather than pure Python code in the fact_ajax function?

redref:
Honestly, I don't know. Now that you said it, it really seems legit.

mterzo:
If the data isn't quite the same as you want in the client the template makes some sense.

redref:
No preference on my side. Let me know if you prefer me to amend those PR's or if you think we can postpone it to do all of these in one pass.

puppetboard/app.py
```diff
@@ -651,104 +656,133 @@ def facts(env):
...
+def fact(env, fact, value):
+    """Fetches the specific fact(/value) from PuppetDB and displays per
     node for which this fact is known.
 
     :param env: Searches for facts in this environment
     :type env: :obj:`string`
     :param fact: Find all facts with this name
     :type fact: :obj:`string`
+    :param fact: Find all facts with this value
+    :type fact: :obj:`string`
```
corey-hammerton: This doc should reference the `value` parameter, not the `fact` parameter.

corey-hammerton:
In the long run if the python can directly output JSON data then that's far less code for this project to maintain.

PR #358, opened 2017-02-10
Moving version to a single place in version.py

mterzo:
Simplify version location.

Bumped version to a development tag.


mterzo:
@redref I wanna bump the version before we pull in the data table changes. Any suggestions on this PR?

PR #359, opened 2017-02-10
Inherit from requirements.txt instead of duplicating

mterzo:
Fixes #358


redref:
I wonder if getting requirements.txt into setup.py / install_requires may be a good idea. Unduplicating some lines.

Found also setup.py / test_require option.

mterzo:
Closing for PR #358

PR #363, opened 2017-02-15
Don't use ranges, use exact releases of dependencies.

mterzo:
Docker build will use pip to install wheels, so that gcc is not
required.


redref:
May be suitable on docker side, but global pinning is a strong choice otherwise.

On my side, I really like mutualized libraries even from PIP. I also think this may not suit rpm/deb release systems really fine.

mterzo:
I'm not sure who's building maintaining the debs/rpm packages.

I think most are installing from puppet-puppetboard. Which may not be the case for those using the openbsd port. 

I'm finding odd behavior and warnings when packages are getting updated because we allow a very loose range.

redref:
Ok. Tested on my side without dependencies modifications, only other fixes.

Seems like related to setuptools called directly instead of pip command line.

PR #367, opened 2017-02-28
[WIP] Simply build requirements.  Dependency management based on

mterzo:
social_app django project.


mterzo:
requesting this later, was added while travis was broken.

PR #375, opened 2017-03-20
Simple requirements

mterzo:
Use pip '-r' in the requirements files and use them to create a simple requirements hierarchy.

Use the new requirements.txt and requirements-test.txt inside of setup.py to define the requirements for each of test and install dependencies.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,12 +1,7 @@
 [tox]
-envlist = py{26,27,35,36}-{default,pinned,unpinned}
+envlist = py{26,27,35,36}
 
 [testenv]
-setenv =
-    TEST_DEPS=True
-    pinned: DEPS_RESOLVE=PINNED
-    unpinned: DEPS_RESOLVE=UNPINNED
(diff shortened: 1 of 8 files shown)


mterzo:
Coveralls is failing now Werkzeug is newer, and the code is not getting executed that handled the older versions.

tox.ini
```diff
@@ -1,12 +1,7 @@
...
+envlist = py{26,27,35,36}
 
 [testenv]
-setenv =
-    TEST_DEPS=True
-    pinned: DEPS_RESOLVE=PINNED
-    unpinned: DEPS_RESOLVE=UNPINNED
 commands=
     py.test --cov=puppetboard --pep8 -v
     py{27,35,36}: bandit -r puppetboard
```
bastelfreak: no py26 here?

mterzo: bandit doesn't work on py26.  it can only scan 27, 35, and 36

PR #384, opened 2017-05-16
Fix dailychart.js when using non-root location

Celant:
When puppetboard is being run from a directory instead of the root of the domain, the daily chart will not work as it tries to load the data from the root. This change makes it load the data from the current working URL, instead of forcing root.

### DIFF
--- a/puppetboard/static/js/dailychart.js
+++ b/puppetboard/static/js/dailychart.js
@@ -1,6 +1,6 @@
 jQuery(function ($) {
   function generateChart(el) {
-    var url = "/daily_reports_chart.json";
+    var url = "daily_reports_chart.json";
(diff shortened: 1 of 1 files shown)


Celant:
Example of issue:

**Puppetboard is located at:**
http://celant.co.uk/puppetboard/
**Puppetboard expects the data to be at:**
http://celant.co.uk/daily_reports_chart.json
**Where the data actually is at:**
http://celant.co.uk/puppetboard/daily_reports_chart.json

kirkins:
Links return 404 or is that the issue?

Celant:
Sorry, those were examples, not actually live. I'll edit that and clarify

mterzo:
thanks!

Issue #365, opened 2017-02-17
Radiator view uses incorrect colours


chris-bmj:
Everywhere else 'Unchanged' nodes are green, 'Changed' are blue. Except in Radiator view where they are the other way round.

PR #388, opened 2017-06-10
Updating radiator CSS to match that of the main CSS.

mterzo:
Changed wording for Pending to Noop. Fixing #365


mterzo:
Old Radiator
old_raditor

New Radiator
new_radiator

PR #389, opened 2017-06-10
Rework "nodes" with datatables.

redref:
This PR will need voxpupuli/pypuppetdb#116 to work.

Worked on this for personal use. Fine on my side, but may need improvements or clarity.

A lot of lines refers to puppetdb 3.X inability to provide a consistent "Node" status on "noop" runs. This might be simplified by dropping 3.X support. By the meantime, this PR contains compatibility functions.

For information, this also include a fix for @phybros latest input (report.metrics == None).

### DIFF
--- a/puppetboard/app.py
+++ b/puppetboard/app.py
@@ -395,7 +395,7 @@ def nodes_ajax(env):
 
     status_query = OrOperator()
     for status_arg in status_args:
-        if status_arg not in ['', '*']:
+        if status_arg not in ['', '*', 'none']:
             arg_query = get_node_status_query(status_arg)
             if arg_query:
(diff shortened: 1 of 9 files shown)


mterzo:
Let's get the 0.3 release out then for 0.4 drop python 2.6 and puppetdb 3

bastelfreak:
@mterzo can you please have a look at this?

yakatz:
If this works, it could be a solution to #567, but in my testing, I get error after error - I suspect it needs an update for some new version of `pypuppetdb`.

gdubicki:
I think that we can safely drop PuppetDB 3.x support. It’s unsupported since 2018, IIRC.

PR #397, opened 2017-07-07
Add offline mode for Semantic UI.

### LINKED ISSUE #304: Offline mode still pulls content from web

mterzo:
Following a similar guide to:
https://github.com/Semantic-Org/Semantic-UI/issues/1521

Using a new endpoint to handle offline mode to host semantic.min.css and
other CSS files.  Need to wrap these in templates to provide URL for
static font files when hosting under a directory.

Fixes #304


bastelfreak:
looks good to me as well, but needs to be rebased.

mterzo:
Its in the mainline. Haven't cut a release yet.

Issue #399, opened 2017-07-10
Nodes sort by catalog or report


kirkins:
I'm on the Nodes page and I'm trying all the sort by column buttons.

Status, and Certname work but Catalog and Report don't work.

Can anyone test on their setup? I want to see if the issue is specific to my setup.
![untitled](https://user-images.githubusercontent.com/9503662/28035469-ba98df4c-6582-11e7-8eac-cb032fbe1ba6.png)![untitled](https://user-images.githubusercontent.com/9503662/28035502-d82b2e0c-6582-11e7-81c1-f5a4ea38a2b7.png)

PR #400, opened 2017-07-11
Remove date classes from nodes page

kirkins:
Related to #399

This change makes sorting by date on the nodes page work.

I noticed the date class seems to trigger the code on line 13 in `Puppetboard / puppetboard / static / js / tables.js`. But other pages, for example Catalogs, have dates that don't use the `.date` class and work.

So I removed the date class and it seems to fix the issue.

I'm wondering if this tables.js code for dates is being used elsewhere, if so is it working, if not should it be removed?


mterzo:
Thanks!

PR #401, opened 2017-07-11
Fix sort by date on nodes and index page

kirkins:
This is a follow up to #400 

I'm removing this code because I can't find it referenced anywhere other than the places I removed it from.

I realize this pull request doesn't add any functionality, but if someone adds a class of date to a `th` element this change will stop the functionality from breaking.

### DIFF
--- a/puppetboard/static/js/tables.js
+++ b/puppetboard/static/js/tables.js
@@ -11,8 +11,9 @@
   }
 
   $('thead th.date').data('sortBy', function(th, td, tablesort) {
-    var tdTime = td.text().replace("-", "");
-    return moment.utc(new Date(tdTime)).unix();
+    var tdTime = new Date(td.text().replace("-", ""));
+    if(isNaN(tdTime)) return 0;
+    else return tdTime;
   });
 
   $('input.filter-table').parent('div').removeClass('hide');
--- a/puppetboard/static/js/tables.js
+++ b/puppetboard/static/js/tables.js
@@ -12,7 +12,7 @@
 
   $('thead th.date').data('sortBy', function(th, td, tablesort) {
     var tdTime = td.text().replace("-", "");
-    return moment.utc(tdTime).unix();
+    return moment.utc(new Date(tdTime)).unix();
   });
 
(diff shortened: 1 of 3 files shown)


kirkins:
Realizing that the sort function isn't working. Instead of deleting this I will go back and try to fix the code.

Example of not working:

untitled

mterzo:
It's used. About 4 months ago I cleaned up the ordering of all the JavaScript.

kirkins:
It should trigger on any table header with class of date. I didn`t see any of those when I looked, except for the ones I removed. 

If you find any with the class I'll be interested to see if it works.

mterzo:
So it's sorting alpha numeric instead of by date?

kirkins:
@mterzo yes I think so. I think it's the case for tables referencing date or time.

kirkins:
I will try to fix it when I go back to work tomorrow if it's still an issue. I don't have it setup at home right now.

kirkins:
Brought back the classes deleted in #400.

Found the javascript was breaking because of the " - " in the date-time string.

kirkins:
This may need another modification, I'm seeing this warning in javascript console log:

> Deprecation warning: moment construction falls back to js Date. This is discouraged and will be removed in upcoming major release. Please refer to https://github.com/moment/moment/issues/1407 for more info.

kirkins:
Sorry for all the commits and then removing. I thought I made a typo in push.

mterzo:
Gonna look at this tomorrow.

kirkins:
@mterzo Have a chance to see if my change works on your end?

larsnaesbye:
So is the sorting correct in upstream as of now?

kirkins:
Without the change sorting of dates is done alphabetically.

So you won't notice this bug unless you have a date based column with different months.

For testing you can modify the dom directly and then resort to see the problem.

larsnaesbye:
OK, I thought this also affected sorting in all other types of numerical lists. I'd like to see lists of integers, like the 'uptime' facts, sorted numerically instead of alphabetically.

larsnaesbye:
@mterzo can we have this merged in?

PR #408, opened 2017-07-26
Preety print multiline diff messages from puppetdb

lkosz:
This PR fixes unreadable diff (and other multiline) messages from puppetdb - it splits message by newline character and show in separate paragraph.


gdubicki:
Closing as #576 that fills the same purpose is merged. Please open a new PR if you feel that there is still something to be done about this, @lkosz .

PR #420, opened 2017-10-17
semantic.min.css not found when OFFLINE_MODE is enabled

jcchak:
Hi!

Just updated Puppetboard to 0.3.0 and I'm having some CSS problems when I leave enabled the offline mode.

Found this in the logs:
`[Tue Oct 17 20:40:21.996344 2017] [wsgi:error] [pid 54588] TemplateNotFound: static/Semantic-UI-2.1.8/semantic.min.css`

And when I checked the HTML code I found this:

> #OFFLINE MODE=True
`<link href="/offline/Semantic-UI-2.1.8/semantic.min.css" rel="stylesheet" />
`
> #OFFLINE_MODE=False
`<link href="/static/Semantic-UI-2.1.8/semantic.min.css" rel="stylesheet" />`

Thanks in advance!

### DIFF
--- a/puppetboard/templates/layout.html
+++ b/puppetboard/templates/layout.html
@@ -13,7 +13,7 @@
       }
     </style>
     <link href='{{ url_for('static', filename='jquery-datatables-1.10.13/dataTables.semanticui.min.css') }}' rel='stylesheet' type='text/css'>
-    <link href="{{ url_for('offline_static', filename='Semantic-UI-2.1.8/semantic.min.css') }}" rel="stylesheet" />
+    <link href="{{ url_for('static', filename='Semantic-UI-2.1.8/semantic.min.css') }}" rel="stylesheet" />
     {% else %}
     <link href='//fonts.googleapis.com/css?family=Open+Sans' rel='stylesheet' type='text/css' />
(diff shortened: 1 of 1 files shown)


mterzo:
Let me look at this. It’s supposed to use a different file. The offline css was modified so it does not get fonts from google.

jcchak:
Thanks Mike for checking!

I tried to locate the file `semantic.min.css` and just found one ocurrence in the path `/usr/lib/python2.7/site-packages/puppetboard/static/Semantic-UI-2.1.8/semantic.min.css`. No traces of a `static/css` folder in `templates`.

I upgraded Puppetboard with pip (`pip install puppetboard --upgrade`)...

https://pypi.python.org/pypi/puppetboard/0.3.0

The tar.gz doesn't have any `static` folder in `templates` neither.

Best regards.

mterzo:
Ha, now that you say that I need to make sure it’s added to the data for in setup.py. Never tested the package build.

larsnaesbye:
I also had to set offline to false to make it work, which was no big deal for me. Would be nice to have fixed eventually, though.

mterzo:
Ok, missed adding the files to the package.  We need to use a templated version of Semantic and google fonts in OFFLINE_MODE so that we point to the offline static instead of the ones that are embedded to the remote.  Closing this PR for #423

Issue #425, opened 2017-10-18
Pie could use a little sugar


larsnaesbye:
The pie chart is real nice, but I could use some aesthetic adjustments, most notably with the mouseover text:
 * There should be some kind of separation between the value and percentage, e.g. by bracketing or even just a space
 * When mousing over, the static labels should disappear so it doesn't display the percentage twice

![puppetboard pie chart](https://user-images.githubusercontent.com/7767079/31720769-c59248b4-b417-11e7-9c75-42839f1cf12d.png)

kirkins:
I can't find any instances on of the pie chart on my board. What page can I find this on?

larsnaesbye:
With default settings, you can see it on 'puppetversion' under Facts, for example.

larsnaesbye:
I just now realised that it uses the C3.js Javascript library to generate charts.

kirkins:
I see I wonder how it is determined which facts show the pie. I have other facts which are simply true/false. You'd think these would show the pie chart too.

larsnaesbye:
It's in your 'default_settings.py' which type to use and which facts to show it on :-)

kirkins:
I see GRAPH_FACTS thx

larsnaesbye:
You can set it to 'bar' or 'donut' too, I think.

larsnaesbye:
Now those examples over at the C3.js site look a lot more sexy: http://c3js.org/samples/chart_pie.html

kirkins:
I like the chart you linked. Is it just a matter of updating to a later version of C3.js?

larsnaesbye:
I think we may have to modify our own CSS too.

larsnaesbye:
That was fast :)

mterzo:
Didn't really have to do anything.  I like the easy ones!

PR #427, opened 2017-10-18
Chart styling

mterzo:
Fixing #425 

screen shot 2017-10-18 at 3 51 20 pm


mterzo:
We were missing the stylesheet.

PR #429, opened 2017-10-19
WIP: button for exporting tables as csv 

emlun007:
Changes are made in layout.html and button libraries were added:

<script type="text/javascript"String src="https://cdnjs.cloudflare.com/ajax/libs/pdfmake/0.1.32/pdfmake.min.js"String></script> <script type="text/javascript"String src="https://cdnjs.cloudflare.com/ajax/libs/pdfmake/0.1.32/vfs_fonts.js"String></script> <script type="text/javascript"String src="https://cdn.datatables.net/v/dt/jszip-2.5.0/dt-1.10.16/b-1.4.2/b-html5-1.4.2/b-print-1.4.2/datatables.min.js"String></script>

### DIFF
--- a/puppetboard/templates/layout.html
+++ b/puppetboard/templates/layout.html
@@ -2,7 +2,7 @@
 <html lang="en">
   <head>
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
-    <title>{{config.PAGE_TITLE}}</title>
+    <title>Puppetboard</title>
     {% if config.OFFLINE_MODE %}
     <style>
       @font-face {
@@ -18,14 +18,18 @@
     <link href='//fonts.googleapis.com/css?family=Open+Sans' rel='stylesheet' type='text/css' />
     <link href="{{ url_for('static', filename='Semantic-UI-2.1.8/semantic.min.css') }}" rel="stylesheet" />
     <link href='//cdnjs.cloudflare.com/ajax/libs/datatables/1.10.13/css/dataTables.semanticui.min.css' rel='stylesheet' type='text/css'>
+    
     {% endif %}
-    <link href="{{ url_for('static', filename='css/c3.min.css') }}" rel="stylesheet" />
     <link href="{{ url_for('static', filename='css/puppetboard.css') }}" rel="stylesheet" />
 
     {% if config.OFFLINE_MODE %}
       <script src="{{ url_for('static', filename='jquery-2.1.1/jquery.min.js') }}"></script>
       <script src="{{ url_for('static', filename='jquery-datatables-1.10.13/jquery.dataTables.min.js') }}"></script>
       <script src="{{ url_for('static', filename='jquery-datatables-1.10.13/dataTables.semanticui.min.js') }}"></script>
+      <script src="{{ url_for('static', filename='DataTables/Buttons-1.4.2/js/dataTables.buttons.min.js') }}"></script>
+      <script src="{{ url_for('static', filename='DataTables/Buttons-1.4.2/js/dataTables.semanticui.min.js') }}"></script>
+      <script src="{{ url_for('static', filename='DataTables/Buttons-1.4.2/js/js/buttons.html5.min.js') }}"></script>
+      
       {% if config.LOCALISE_TIMESTAMP %}
       <script src="{{ url_for('static', filename='moment.js-2.7.0/moment.min.js') }}"></script>
       {% endif %}
@@ -46,7 +50,8 @@
     <script src="{{url_for('static', filename='js/d3.min.js')}}"></script>
     <script src="{{url_for('static', filename='js/c3.min.js')}}"></script>
     <script src="{{url_for('static',
-                  filename='jquery-tablesort-v.0.0.11/jquery.tablesort.min.js')}}"></script>
+    filename='jquery-tablesorter-v.2.0.5/jquery.tablesorter.min.js')}}"></script>
+
     {% block script %} {% endblock script %}
     <script type="text/javascript">
     {% block javascript %} {% endblock javascript %}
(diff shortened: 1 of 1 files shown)


larsnaesbye:
@mterzo : Have you taken a look at this and can you comment on what is needed for it to be merged.

mterzo:
CI is failing on the comment I made earlier. There are a bunch of JavaScript files added for local mode but I don’t see those files in the commits. 

There was a major refactor of moving all the JavaScript to cacheable files instead of having all the JavaScript in HEAD of the html. We need make sure we don’t regress there.

emlun007:
@mterzo ,  The javascripts that I added were for another commit where I am trying to solve the issue with the table sorting, so I will just remove the parts of the code that are not related to save button

puppetboard/templates/layout.html
```diff
@@ -45,8 +47,9 @@
     <script src="{{ url_for('static', filename='js/scroll.top.js') }}"></script>
     <script src="{{url_for('static', filename='js/d3.min.js')}}"></script>
     <script src="{{url_for('static', filename='js/c3.min.js')}}"></script>
-    <script src="{{url_for('static',
-                  filename='jquery-tablesort-v.0.0.11/jquery.tablesort.min.js')}}"></script>
```
emlun007: Lines 48 and 50 are identical. No changes

puppetboard/templates/layout.html
```diff
@@ -2,7 +2,9 @@
 <html lang="en">
   <head>
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
-    <title>{{config.PAGE_TITLE}}</title>
+    <title>Puppetboard</title>
```
kirkins: You shouldn't change this line in the commit

puppetboard/templates/layout.html
```diff
@@ -2,7 +2,9 @@
...
+    <button id="export" style="clear: right;width:127px;height:50px;color:#0a214a;box-shadow: 4px 4px #0a214a;margin-top: 5 px; margin-left: 5 px">Export data as csv</button>
```
kirkins: You should move the inline style to the CSS block so it stays organized.

puppetboard/templates/layout.html
```diff
@@ -56,8 +60,68 @@
           $.getScript('{{url_for('static', filename='js/tables.js')}}')
           {% block onload_script %} {% endblock onload_script %}
         })
+
+    
+      $('#export').click(function() {
```
kirkins: Try to use proper indentation.

puppetboard/templates/layout.html
```diff
@@ -56,8 +60,68 @@
...
+      var titles = [];
+      var data = [];
+
+  $('.dataTable th').each(function() {
```
kirkins: Moving my comment about indentation, as this where you stopped indenting.

If line 67 is indented and no block has ended this should also be indented. 

Also another small thing lines 66 and 67 should have a 2 space indent as children of the function starting on 65.

kirkins:
I see the button and when I press it a file downloads.

But the file is empty. I don't see any errors in the javascript console log.

Does the csv contain data on your setup?

kirkins:
Other thoughts

1. Remove the favicon, it's cute but not related to this PR.
2. Nice job with the drop shadow but you will have to fix the design on this. I recommend integrating the button into the existing navigation bar.

Here is a preview of the button for those who haven't run the code locally:

untitled

emlun007:
The _macros.html is a different PR, same for favicon, I have no idea how the files got in this PR. The only change is in layout.html.

emlun007:
@kirkins , I updated layout.html with the version I am using. It works for chrome

kirkins:
@emlun007 common mistake. You're using your master branch for this PR. That means the PR will reflect the state of your master branch.

Currently that branch has changes for multiple intended PRs.

You'll want to make a separate branch for each PR.

puppetboard/templates/layout.html
```diff
@@ -56,6 +57,64 @@
           $.getScript('{{url_for('static', filename='js/tables.js')}}')
           {% block onload_script %} {% endblock onload_script %}
         })
+       
+      $('#export_data_button').click(function() {
+        var titles = [];
+        var data = [];
+
+        $('.dataTable th').each(function() {
```
kirkins: The table doesn't have this class name on any of the pages I'm looking at.

What page did you test this on?

kirkins: Ok I see it works perfect on the Inventory page.

Do you only intend to use it with inventory? Or do you want it to work on other pages too?

emlun007: No, it works for all pages. I have tested on all facts, and it works there too

emlun007: What is the browser? I have tested with IE and sometimes it doesn't work there

kirkins: It seems to work on all fact pages for me too.

It doesn't work on Overview, Nodes, Facts, Reports, Metrics, Catalogs (kind of works), or Query.

emlun007: This is because the data in Nodes,Facts ( try opening a fact, e.g. "ip adresses" there it works), reports is not initialized from dataTable, so there are no headers and data cells. So this code will only work on any data from dataTable

kirkins:
For style a better option might be just adding a CSV button to the menu:

untitled

`<a id="export_data_button" class="item" href="#">CSV</a>`

Then maybe add some logic to hide the button on pages it doesn't work with.

kirkins:
I agree it is a nice feature. I don't think it can be released as is because people will see it on the first page, try, and think it's not working.

Also follow up on my menu suggestion. It might be a bit tricky because currently the menu is generated with:

```

      {%- for endpoint, caption in [
      ('index', 'Overview'),
      ('nodes', 'Nodes'),
      ('facts', 'Facts'),
      ('reports', 'Reports'),
      ('metrics', 'Metrics'),
      ('inventory', 'Inventory'),
      ('catalogs', 'Catalogs'),
      ('radiator', 'Radiator'),
      ('query', 'Query'),
] %}
```

Which doesn't seem to be able to take a value like `('#', 'CSV'),`

emlun007:
I have tried to modify default_settings.py that defines the menu, but if you use anything else than 'fact name', 'fact', it gives 505 error so I don't think you can change the menu structure.

The button should work on the all th and td rows. The table uses .dataTable th and .dataTable td as input  and it might be the case for facts, but I am unsure how nodes table is populated. I'm not too good with template engines so any help is appreciated.

gdubicki:
Are you interested in getting back to this PR, rebasing and making it work @emlun007?

PR #435, opened 2017-11-07
Add sorting IP adresses in tables

emlun007:
I have added code in tables.js that will sort the IP adresses in inventory table. Right now, the function does not work properly, most likely because I don't use the right name of the variable with IP Adresses

### DIFF
--- a/puppetboard/static/js/tables.js
+++ b/puppetboard/static/js/tables.js
@@ -16,6 +16,17 @@
     else return tdTime;
   });
 
+    $('thead th.IP Address').data('myIpArray', myIpArray) // save IP adresses from th. IP Addresses (unsure about the name) in myIpArray
+  myIpArray.sort(function(a,b){ // sort array in myIpArray
+    aa = getIP(a).split(".");
+    bb = getIP(b).split(".");
+  
+    for (var i=0, n=Math.max(aa.length, bb.length); i<n; i++) {
+        if (aa[i] !== bb[i]) return aa[i] - bb[i];
+    }
+    return 0;
+  });
+  
   $('input.filter-table').parent('div').removeClass('hide');
 
   $("input.filter-table").on("keyup", function(e) {


puppetboard/static/js/tables.js
```diff
@@ -16,6 +16,17 @@
     else return tdTime;
   });
 
+    $('thead th.IP Address').data('myIpArray', myIpArray) // save IP adresses from th. IP Addresses (unsure about the name) in myIpArray
```
kirkins: It looks like you're trying to select `th.IP Address`, following the pattern for `.date` sort.

Are you sure this class exists on any of the table headers? If there isn't a class name identifying `IP Address` in the html, you'd have to add one for this to work.

If you do add it use an underscore `ip_address` as CSS classes can't have spaces.

kirkins: This is right clicking in chrome on a date header and clicking inspect. It shows `class="date"`.

When I do the same on anything with an IP I don't any unique class.

untitled

emlun007: The reason I tried IP Address as a table header is when I import the headers from the table I get something like:
"Hostname,IP Address,OS,Architecture,Kernel Version,Reboot required,Puppet Version"
So IP Address must be a header so I wrote th.IP Address

kirkins: For the Inventory page I can select the `th` for IP Address with this:

    $("table").find("[aria-label^='IP Address']")

It says find in table where `aria-label` starts with 'IP Address'.

emlun007: I can see them as headers when I try the sorting on inventory. 
ip address

kirkins: I see where you were coming from, but the syntax isn't correct. Just run your selector in the console tag on chrome and you will see it returns 0 items.

untitled

emlun007: I might need to create a class, but then how should I identify it if only reference there is is the aria-label^='IP Address'?

emlun007:
Update: Now the ip sorter finally works 😃 

All you have to do is to write column number in _macros.html which is set to 2nd column by default:

{ type: 'ip-address', targets: 1 }

kirkins:
I'm testing this now.

Everything seems to be working. On the Inventory page X.X.0.10 is now considered greater than X.X.0.2 for sorting.

Good work @emlun007

puppetboard/templates/_macros.html
```diff
@@ -56,6 +56,9 @@
     "search": {"regex": true},
     // Default sort
     "order": [[ 0, "desc" ]],
+    //Sort ipadresses
```
kirkins: NIT: Typo in address and add space to match with line 57.   `// Sort ip address`

emlun007:
@kirkins , I have added code for sorting uptime (xxx days) and memory (xxx GiB). The code parses strings to integer (9.99 GiB = 9) and checks in the result is NaN. If the sourse is NaN (e.g. ip address), the original data is shown, if the source is not NaN, it is parsed to integer

puppetboard/templates/_macros.html
```diff
@@ -56,6 +56,21 @@
...
+
+    {"targets": -1,
+    "render": function ( data, type, row )
+     {if (isNaN(parseInt(data)))
+      return (data)
+      else
+     return parseInt(data)}
+     },
+    ],
+    // Custom options
```
vchepkov: Duplicate comment

larsnaesbye:
Can this be merged, including @vchepkov 's changes? :-)

bastelfreak:
Sadly, it cannot be merged. This branch has merge conflicts. You can checkout the branch,  resolve the conflicts and submit it as a new PR. I'm happy to review and merge it afterwards.

Issue #404, opened 2017-07-13
Padding on node table


kirkins:
![untitled](https://user-images.githubusercontent.com/9503662/28172929-f51fb264-67ba-11e7-86f1-3eb32f9db10b.png)

I found changing this value was an option to fix the style:

![untitled](https://user-images.githubusercontent.com/9503662/28173078-5debd39a-67bb-11e7-988e-fcffcedc5520.png)

Here is what it looks like if I change the first number in decimal from `7` to `4`:

![untitled](https://user-images.githubusercontent.com/9503662/28173128-80e0697e-67bb-11e7-8aba-e18c969fa5f7.png)

It seems the origin is defined by Semantic UI so I'm hesitant to change it. But I think it looks better when it all fits on a single line. Does the style issue occur for anyone else?

kirkins:
Then again every user will have their own screen-size. 

So even if the padding was changed it's sure to flow to next line in some situations based on screen-size.

emlun007:
I have to use 70% zoom to be able to see what is on the page :D so it would be useful to make the button smaller

kirkins:
@emlun007 that sucks. Do you know how many pixels across your browser is?

I posted a fix that seems to help on screens between 1200px-1500px. 

On my setup at least the problem was the Certname column was too wide.

gdubicki:
I assume that this is fixed by #436 and therefore closing this issue. Please reopen if it is not.

PR #436, opened 2017-11-07
Make 320px the max width for columns #404

kirkins:
This pull request helps with a layout issue I pointed out in #404. The Certname column is sometimes too long which makes the buttons display not as intended. 

In particular I found this helps on screens 1200px-1500px when cert-names are long.


kirkins:
**Without:**
untitled

**With:**
untitled

emlun007:
Maybe make it even 300? I still seems too long

Issue #424, opened 2017-10-18
Daily charts missing on node page.


mterzo:
Request for daily charts JSON is missing.

```
10.0.0.17 - - [18/Oct/2017:04:53:26 +0000] "GET /node/docker00.private.terzo.org HTTP/1.1" 200 8849 "http://puppetboard.internal.terzo.org/reports" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_13) AppleWebKit/604.1.38 (KHTML, like Gecko) Version/11.0 Safari/604.1.38"
ERROR:pypuppetdb.api:{
  "error" : "No information is known about node daily_reports_chart.json"
} puppetdb:8080 over HTTP.
ERROR:puppetboard.utils:404 Client Error: Not Found for url: http://puppetdb:8080/pdb/query/v4/nodes/daily_reports_chart.json
```

smboucher:
Any additional info that would help resolve this one?

mterzo:
The daily charts JS relies on the fact that it uses relative pathing to get the envto pass on the request for the data. When rendering the node page  the request from the browser to get the JS file is passing it to flask under the node directory causing puppetboard to generate a puppetdb query.

octomike:
So basically, this is just a URL config issue.
This change fixes daily charts for nodes, **but** it might interfere with webserver integration.

```diff
diff --git a/puppetboard/static/js/dailychart.js b/puppetboard/static/js/dailychart.js
index f68d428..a14dca2 100644
--- a/puppetboard/static/js/dailychart.js
+++ b/puppetboard/static/js/dailychart.js
@@ -1,6 +1,6 @@
 jQuery(function ($) {
   function generateChart(el) {
-    var url = "daily_reports_chart.json";
+    var url = "/daily_reports_chart.json";
     var certname = $(el).attr('data-certname');
     if (typeof certname !== typeof undefined && certname !== false) {
       url = url + "?certname=" + certname;
```

Is there some equivalent of `url_for` to get the site root URL in this case?

Otherwise adding a 3rd and 4th route `@app.route('/<node>/<env>/daily_reports_chart.json'` would solve this, but then there is no simple way to differentiate between `/<node>/daily_reports_chart.json` and `/<env>/daily_reports_chart.json` anymore and this becomes a design issue.

bastelfreak:
This should be fixed since https://github.com/voxpupuli/puppetboard/pull/438 got merged?

PR #438, opened 2017-11-07
Fix problem loading daily chart on node page

kirkins:
Fix for #424 by referencing base site url plus daily report location.

### DIFF
--- a/puppetboard/static/js/dailychart.js
+++ b/puppetboard/static/js/dailychart.js
@@ -1,6 +1,8 @@
 jQuery(function ($) {
   function generateChart(el) {
-    var url = window.location.origin + "/daily_reports_chart.json";
+    var url = window.location.protocol + "//" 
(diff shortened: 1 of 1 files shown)


kirkins:
**Daily node page with change:**

untitled

**Chart on front-page still working:**

untitled

mterzo:
Can we trust that DOM value for all browsers?

kirkins:
@mterzo good call, I think there are some older browsers that will have trouble with `window.location.origin` I'm going to add a polyfill to handle that.

kirkins:
Updated for IE, with solution from here https://stackoverflow.com/a/25495161/773263

mterzo:
I was gonna keep this simple and move dailychart to the templates and generate the base url.

bastelfreak:
@mterzo can you please take a look at this again?

Issue #314, opened 2016-11-04
Sort order forgotten when page refreshes


Flupsy:
If a non-default sort order is selected, the order is reset when the page refreshes.
We're using version 0.2.0.

corey-hammerton:
Table sorting is done with JavaScript which cannot remember anything beyond page reloads.

nikolaik:
It's possible to put the sort order in a URL query param (onClick) and then pick up the param and sort with the same JS (on DOMContentLoaded).

PR #439, opened 2017-11-07
Record sort order in URL for auto-refreshing

kirkins:
This is a code snippet that allows the sort order to be controlled via a URL parameter set by clicking on a table header. #314

### DIFF
--- a/puppetboard/static/js/lists.js
+++ b/puppetboard/static/js/lists.js
@@ -40,3 +40,26 @@
   });
 
 }).call(this);
+
+// Add function to easily get URL params
+$.urlParam = function(name){
+  var results = new RegExp('[\?&]' + name + '=([^&#]*)').exec(window.location.href);
+  if (results==null){
+    return null;
+  }
+  else{
+    return decodeURI(results[1]) || 0;
+  }
+}
+
+// On startup if url param sort is set replace default sort
+if($.urlParam('sort')) {
+  $('.default-sort').removeClass('default-sort');
+  $("th:contains('"+$.urlParam('sort')+"')").addClass('default-sort');
+}
+
+// When a table header is clicked update the URL without refreshing page
+$( "th" ).click(function() {
+  var newUrl = location.origin + location.pathname + "?sort=" + $(this).text();
+  history.pushState({}, null, newUrl);
+});


kirkins:
@mterzo let me know if this is Ok so far. If so I will add a second parameter for ascending vs descending.

mterzo:
We shouldn’t have to post back to the server on this. I haven’t had a chance to look but most JS sort library’s have a display to show and an actual value. For IPs we just need to conver them to their interfered value.

mterzo:
I’m also looking into the larger scale installations where we are causing too many queries back to PuppetDB. If we don’t have to requery data we already have that would be helpful. Of course if we are paging the results the sort is on the result set not the entire data set.

kirkins:
@mterzo this doesn't post back to the server I'm using an HTML5 feature that lets you update the URL without querying the website.

https://developer.mozilla.org/en-US/docs/Web/API/History_API#The_pushState()_method

Also this isn't related to the IP sorting. It is related to the fact that Puppetboard automatically refreshes every X seconds and forgets which column the user was sorting by.

puppetboard/static/js/lists.js
```diff
@@ -40,3 +40,26 @@
...
+
+// On startup if url param sort is set replace default sort
+if($.urlParam("sort")) {
+  $(".default-sort").removeClass("default-sort");
+  $("th:contains('"+$.urlParam("sort")+"')").addClass("default-sort");
+}
+
+// When a table header is clicked update the URL without refreshing page
+$("th").click(function() {
+  var newUrl = location.origin + location.pathname + "?sort=" + $(this).text();
```
kirkins: Would have to replace the above `location.origin` with polyfill too.

puppetboard/static/js/lists.js
```diff
@@ -40,3 +40,26 @@
...
+  history.pushState({}, null, newUrl);
```
kirkins: @mterzo this probably isn't compatible with older browsers. I'll do some more research this week.

emlun007: It works for Internet explorer 👍

bastelfreak:
@mterzo can you please have a look at this?

raphink:
This works fine for me, but it only records the sort key, not the order (`desc` or `asc`).

Issue #406, opened 2017-07-13
Repeat javascript on main page


kirkins:
- `c3.min.js`, included 4 times
- `d3.min.js`, included 3 times
- `dailychart.js` included 2 times

![untitled](https://user-images.githubusercontent.com/9503662/28177221-1c34e77c-67c8-11e7-9d32-737930023de1.png)

kirkins:
I noticed that on the index removing this from `layout.html`:

    {% block head %} {% endblock head %}

Removes lines 46-52 and nothing else. Do you know where the head block defined?

mterzo:
It's defined here: https://github.com/voxpupuli/puppetboard/blob/3867e1aa77e2b24eae56f796df69777e1c955988/puppetboard/templates/layout.html#L60  We still want to keep the block, just need to remove the these from the index.html template.

gdubicki:
As #441 is merged I assume that we can close this issue. Please reopen if needed.

PR #441, opened 2017-11-10
remove duplicate c3.js/d3.js links

kirkins:
Removing extra links reported in #406 

Even with this change there seems to be 2 of both `d3.js` and `c3.js` on index.

### DIFF
--- a/puppetboard/templates/index.html
+++ b/puppetboard/templates/index.html
@@ -6,6 +6,7 @@
 {% endif %}
 {% if config.DAILY_REPORTS_CHART_ENABLED %}
 {% block script %}
+<script src="{{url_for('static', filename='js/c3.min.js')}}"></script>
 <script src="{{url_for('static', filename='js/d3.min.js')}}"></script>
 <script src="{{url_for('static', filename='js/dailychart.js')}}"></script>
 {% endblock script %}
--- a/puppetboard/templates/index.html
+++ b/puppetboard/templates/index.html
@@ -6,8 +6,8 @@
 {% endif %}
 {% if config.DAILY_REPORTS_CHART_ENABLED %}
 {% block script %}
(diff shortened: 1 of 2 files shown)


puppetboard/templates/index.html
```diff
@@ -7,7 +7,6 @@
 {% if config.DAILY_REPORTS_CHART_ENABLED %}
 {% block script %}
 <script src="{{url_for('static', filename='js/d3.min.js')}}"></script>
-<script src="{{url_for('static', filename='js/c3.min.js')}}"></script>
```
kirkins: I removed this because on line 4-6 we have:

```
{% if config.DAILY_REPORTS_CHART_ENABLED %}
<link href="{{ url_for('static', filename='css/c3.min.css') }}" rel="stylesheet">
{% endif %}
```

Should `d3.js` be in the block as well, or is it used elsewhere?

kirkins:
@mterzo I might have read what you said wrong.

Is it that `c3.js` and `d3.js` should only appear in `layout.html` and not `index.html`?

puppetboard/templates/index.html
kirkins: Still waking up I shouldn't have removed this was mixing js and css up.

kirkins:
I think it makes more sense that `c3.js` and `d3.js` are in `index.html` and `node.html` instead of `layout.html`.

It seems those scripts are only applicable on those pages so shouldn't be included elsewhere.

kirkins:
Wait the script also needs to be on the fact.html page

larsnaesbye:
Thinking this should be a straightforward merge(?) @mterzo , if you don't have the time to oversee merging of pull requests and/or releases, could someone else be assigned to assist?

othalla:
good idea :)

PR #444, opened 2017-12-02
Fix: default table sort was always ascending

octomike:
An ascending sort order is always triggered when then target index is different
from the tablesort index. This was always true on a page (re)load, because
index is initialized with null.  By setting the instance' index to the
default-sort column index we assure that the direction parameter to sort() is
used.

See https://github.com/kylefox/jquery-tablesort/blob/master/jquery.tablesort.js#L42

### DIFF
--- a/puppetboard/static/js/tables.js
+++ b/puppetboard/static/js/tables.js
@@ -12,10 +12,11 @@
     else return tdTime;
   });
 
-  if ($('th.default-sort').data()) {
+  var thDefaultSort = $('th.default-sort');
+  if (thDefaultSort.data()) {
     var tablesort = $('table.sortable').tablesort().data('tablesort');
-    tablesort.index = $('th.default-sort').index();
-    tablesort.sort($("th.default-sort"), 'desc');
+    tablesort.index = thDefaultSort.index();
+    tablesort.sort(thDefaultSort, 'desc');
   }
 
 
--- a/puppetboard/static/js/tables.js
+++ b/puppetboard/static/js/tables.js
@@ -7,7 +7,11 @@
   $(function() {});
 
   if ($('th.default-sort').data()) {
-    $('table.sortable').tablesort().data('tablesort').sort($("th.default-sort"), "desc");
+    $.tablesort.DEBUG = false;
+    var tablesort = $('table.sortable').tablesort().data('tablesort');
+    // explicitly set the current index so we don't get default sort (asc)
+    tablesort.index = $('th.default-sort').index();
+    tablesort.sort($("th.default-sort"), 'desc');
   }
 
   $('thead th.date').data('sortBy', function(th, td, tablesort) {
--- a/puppetboard/static/js/tables.js
+++ b/puppetboard/static/js/tables.js
@@ -6,19 +6,18 @@
 
   $(function() {});
 
-  if ($('th.default-sort').data()) {
-    $.tablesort.DEBUG = false;
-    var tablesort = $('table.sortable').tablesort().data('tablesort');
-    // explicitly set the current index so we don't get default sort (asc)
-    tablesort.index = $('th.default-sort').index();
-    tablesort.sort($("th.default-sort"), 'desc');
-  }
-
   $('thead th.date').data('sortBy', function(th, td, tablesort) {
     var tdTime = new Date(td.text().replace("-", ""));
     if(isNaN(tdTime)) return 0;
     else return tdTime;
   });
+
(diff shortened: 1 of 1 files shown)


emlun007:
Tablesort will not work properly with DataTables that is initialized in _macros.html.

octomike:
I'm not trying to save anything.

Maybe it's just me, but right now the default sorting on the index page is broken. I assume the only logical order is descending by report timestamp, right? And this PR should address this.

kirkins:
It should or it does? @octomike

octomike:
Not for me...

This is current master:

2017-12-20-154414_688x1108_scrot

On month changes this would look even worse, because the ordering is not just in the wrong direction, it's also broken.

This is the feature branch:

2017-12-20-154927_630x848_scrot

puppetboard/static/js/tables.js
```diff
@@ -6,16 +6,21 @@
...
-  }
-
   $('thead th.date').data('sortBy', function(th, td, tablesort) {
     var tdTime = new Date(td.text().replace("-", ""));
     if(isNaN(tdTime)) return 0;
     else return tdTime;
   });
 
+  if ($('th.default-sort').data()) {
+    $.tablesort.DEBUG = false;
```
kirkins: Is this set to true by default?

puppetboard/static/js/tables.js
```diff
@@ -6,16 +6,21 @@
...
+    var tablesort = $('table.sortable').tablesort().data('tablesort');
+    // explicitly set the current index so we don't get default sort (asc)
```
kirkins: This comment is relevant to the commit but I don't think it's necessary in the source.

You should change as little as possible. Don't move line 9 to line 15 if you don't have to.

The commit should simply remove line 10 and replace it with:

```
var tablesort = $('table.sortable').tablesort().data('tablesort');
ablesort.index = $('th.default-sort').index();
tablesort.sort($("th.default-sort"), 'desc');
```

kirkins:
As for the fix in end behavior. I can confirm that this change fixes the default sorting when you first launch a page.

Nice work 👍

puppetboard/static/js/tables.js
octomike: The reordering of the two calls is necessary unfortunately.

As for the annotations, I just removed them and force pushed.

octomike:
Meh, I force pushed and I guess that deleted the review comments. Sorry about that.
I removed the unnecessary line and the comment (although I would find that helpful as a newcomer). 

The reordering of the function calls are necessary for the fix, however.

octomike:
Are there still issues with the proposed fix?

octomike:
Can this be merged?

puppetboard/static/js/tables.js
```diff
@@ -6,16 +6,19 @@
...
     var tdTime = new Date(td.text().replace("-", ""));
     if(isNaN(tdTime)) return 0;
     else return tdTime;
   });
 
+  if ($('th.default-sort').data()) {
+    var tablesort = $('table.sortable').tablesort().data('tablesort');
+    tablesort.index = $('th.default-sort').index();
+    tablesort.sort($("th.default-sort"), 'desc');
+  }
```
runejuhl: IMO `$('th.default-sort')` should be assigned to a variable to avoid looking up the same selector three times; I imagine that lookup could be fairly slow.

larsnaesbye:
As @octomike said, can't this be merged?

Then the changes requested by @runejuhl, the latter could be a different PR.

runejuhl:
> As @octomike said, can't this be merged?
> 
> Then the changes requested by @runejuhl could be a different PR.

I'd much rather fix issues beforehand -- my experience is that if we don't fix it now, it'll never be fixed, and for minor things like this it's doubly true.

@larsnaesbye care to give it a spin and see if everything is alright? I've rebased @octomike's work onto master, done some linting and fixed the requested change. Find it here: https://github.com/runejuhl/puppetboard/tree/pr/444

larsnaesbye:
I say go for a merge.

puppetboard/static/js/tables.js
raphink, line 19: Agreed @runejuhl. Maybe something like:

```suggestion
var thDefaultSort = $('th.default-sort');
  if (thDefaultSort.data()) {
    var tablesort = $('table.sortable').tablesort().data('tablesort');
    tablesort.index = thDefaultSort.index();
    tablesort.sort(thDefaultSort, 'desc');
  }
```

PR #457, opened 2018-03-21
Allow to force the PuppetDB connection protocol

### LINKED ISSUE #456: Allow to force the protocol to connect to PuppetDB
When the `PUPPETDB_CERT` and `PUPPETDB_KEY` are not set, the Pypuppetdb library will automatically choose 'http' as the protocol to be used for the connection to PuppetDB.

It seems to me that as of now is not possible to configure PuppetBoard to connect to PuppetDB via HTTPS without a key/cert pair.
I think it's a quick and easy feature to add, so I'm sending a PR for it, if you'll deem acceptable.

volans-:
* Closes #456


mterzo:
Thanks.

PR #462, opened 2018-04-13
Fix documentation error regarding ENABLE_CATALOG

rkcpi:
True or False are expected (python booleans start with capital letters)


bastelfreak:
Thanks @rkcpi !

PR #465, opened 2018-04-21
Adding basic health check endpoint

RobReus:
The endpoint /status can be used to determine if puppetboard is running or not.
Reason for needing an endpoint like this is because if the puppetdb has no
data yet (due to a fresh install for example), it will return a 404 on all pages.
This will throw off health checks for example docker or the AWS ALB.

The health check endpoint can later be expanded by puppetdb health as soon
as this functionality has been merged in the pypuppetdb module.


wyardley:
This seems simple and sane to me; thanks for the contribution.

PR #467, opened 2018-04-28
Query using producer_timestamp index vs. start_time

### LINKED ISSUE #466: Improve daily chart query performance
Full details (including sample queries and times) are in a ticket that I had filed with PuppetDB as well: [PDB-3910](https://tickets.puppetlabs.com/projects/PDB/issues/PDB-3910).

We are currently using the latest version of Puppetserver (5.3.0), Puppet (5.5.0), and PuppetDB (5.2.1), and Puppetboard (0.3.0) in an environment with approximately 600 nodes, which have agent running in noop mode splayed every 60 min, and we are retaining reports on all nodes for 2 years.  Over time, this has obviously increased the size of the puppetdb database, specifically the reports table.

Over time, we found some of the daily chart queries taking upwards of 50-60s each due to full table search without an index on the `start_time` column.  I temporarily added one to our database and queries dropped immediately to ~30-40ms.  Initially, I brought this up to the folks at Puppet, and asked for an improvement on PuppetDB by adding an index to the `start_time` column.  They suggested to change the query to use an existing column (`producer_timestamp`) with an index that might be better suited for querying reports.

Report wireform spec: https://puppet.com/docs/puppetdb/4.2/api/wire_format/report_format_v5.html
```
"start_time" is the time on the agent at which the Puppet run began; see more details about the datetime format below.
```

```
"producer_timestamp" is the time of catalog submission from the Puppet master to PuppetDB. This field is currently populated by the Puppet master. See more details about the datetime format below.
```

jplindquist:
Resolves #466

### DIFF
--- a/puppetboard/dailychart.py
+++ b/puppetboard/dailychart.py
@@ -30,8 +30,8 @@ def _build_query(env, start, end, certname=None):
     query.add_field(FunctionOperator('count'))
     query.add_field('status')
     subquery = AndOperator()
-    subquery.add(GreaterEqualOperator('start_time', start))
-    subquery.add(LessOperator('start_time', end))
+    subquery.add(GreaterEqualOperator('producer_timestamp', start))
+    subquery.add(LessOperator('producer_timestamp', end))
     if certname is not None:
         subquery.add(EqualsOperator('certname', certname))
     if env != '*':


bastelfreak:
Hi @jplindquist, thanks for the PR! Do you know since when those fields are available in Puppetdb? (Do we need to label it as breaking change)

jplindquist:
Thanks for the review @bastelfreak.  Looks like migrations were added for this to support upgrades from 2.3.2.  I believe the minimum requirement is PuppetDB 3, so this shouldn't be a breaking change if I'm not mistaken.

ref:

- https://github.com/puppetlabs/puppetdb/commit/1ecca88b530385c74c7607b7248336de46190e64#diff-a744d7c2997eabcde50794a2f5a7d3d5R1025

-https://github.com/puppetlabs/puppetdb/commit/1ecca88b530385c74c7607b7248336de46190e64#diff-a744d7c2997eabcde50794a2f5a7d3d5R806

bastelfreak:
Thanks!

Issue #475, opened 2018-07-11
Allow to configure which PuppetDB endpoints are allowed in the /query page


volans-:
Right now there is just a global switch to enable or disable the `/query` endpoint: `ENABLE_QUERY`.
It would be nice to have an easy way to allow only a sub-selection of the available PuppetDB endpoints.
For example the catalog endpoint might return secrets from the host's catalog and this could be considered unsafe and to be forbidden in a given environment. But at the same time it would be very useful to allow fact queries. And this is my use case.

An additional configuration parameter that allow to specify the list of enabled PuppetDB endpoints that could be queried should be easy enough to add, increasing a lot Puppetboard flexibility.

PR #477, opened 2018-07-18
Allow to configure enabled query endpoints

volans-:
Allow to configure which PuppetDB endpoints are allowed in the `/query`
page, adding the `ENABLED_QUERY_ENDPOINTS` configuration variable, that
if set restricts the allowed endpoints to query. If not set the previous
behaviour is kept, allowing all PuppetDB endpoints to be queried, if the
`ENABLE_QUERY` configuration is `True`.

Closes: #475

### DIFF
--- a/puppetboard/app.py
+++ b/puppetboard/app.py
@@ -19,7 +19,7 @@ from jinja2.utils import contextfunction
 
 from pypuppetdb.QueryBuilder import *
 
-from puppetboard.forms import QueryForm
+from puppetboard.forms import ENABLED_QUERY_ENDPOINTS, QueryForm
 from puppetboard.utils import (get_or_abort, yield_or_stop,
                                get_db_version)
 from puppetboard.dailychart import get_daily_reports_chart
@@ -49,6 +49,7 @@ CATALOGS_COLUMNS = [
     {'attr': 'catalog_timestamp', 'name': 'Compile Time'},
     {'attr': 'form', 'name': 'Compare'},
 ]
+
 
 app = get_app()
 graph_facts = app.config['GRAPH_FACTS']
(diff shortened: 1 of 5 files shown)


volans-:
@alexjfisher Hi, so nice to see some activity in Puppetboard, thanks a lot!
I was wondering if this PR could be evaluated and, in that case, before rebasing and resolving the conflicts, I'd like to know if I should drop the Python 2.6 fix present in the second commit given that 2.6 was dropped from the test matrix.

volans-:
I've rebased the branch to the current HEAD of master resolving the conflicts (was just he README migration from `rst` to `md` format).
I've also took the liberty to remove the second commit that was fixing the compatibility with Python 2.6 given that its support was dropped in the meanwhile.
Already squashed in a single commit, ready for final review.

puppetboard/app.py
```diff
@@ -758,36 +759,42 @@ def query(env):
...
+        'csrf_context': session})
+    if form.validate_on_submit():
+        if form.endpoints.data not in ENABLED_QUERY_ENDPOINTS:
+            log.warn('Access to query endpoint %s disabled by administrator.',
+                     form.endpoints.data)
+            abort(403)
+
+        if form.endpoints.data == 'pql':
+            query = form.query.data
+        elif form.query.data[0] == '[':
```
runejuhl: Should this perhaps be `form.query.data.strip[0]` instead to correctly handle a query with a leading newline/space? I imagine that could happen pretty easily with a textarea field and copy-paste.

volans-: @runejuhl thanks for the review. This was pre-existing code (see line 771 on current code), but I'm happy pass it through `strip()` if there is an agreement.

runejuhl: > /me toggles the unified diff setting

Sorry about that, you're right of course.

bastelfreak: I think this can stay as it is. But another PR that makes the existing code more beautiful is always highly appreciated.

alexjfisher:
@volans- Thanks for the PR!
@runejuhl Thanks for the review!

alexjfisher:
oh and @bastelfreak, thanks for merging and sorting out all the travis related stuff over the weekend! :)

PR #480, opened 2018-10-01
CommonMark >= 0.8.x preparation

covidium:
As commonmark package has been renamed to lowercase and removed symlink starting v0.8.1 it will be a good point to import the module directly to lowercase in order to remove the try: block in the future when using a later version of this package:
https://github.com/rtfd/CommonMark-py/releases


larsnaesbye:
@covidium  - can you try rebasing to see if the checks pass now?

smortex:
I just committed a simplified version of this to fix the FreeBSD port of pupperboard.  This would be awesome to merge the fix.  In order to help with this, I rebased this on top of master and opened another PR: #497

PR #497, opened 2019-03-21
CommonMark >= 0.8.x preparation (rebased)

smortex:
This is a follow-up to #480, I cherry-picked the author's commit on top of master to prove that the test suite pass with these changes (except the Allowed failure that already fails with master).

Feel free to merge this PR or the original one; and close the other one at your convenience

Fixes #480


bastelfreak:
Thanks for the PR!

PR #503, opened 2019-05-26
enable pypi releases through travis

bastelfreak:


.travis.yml
```diff
@@ -1,21 +1,50 @@
 dist: xenial
 language: python
-
 python:
-  - "2.7"
-  - "3.5"
-  - "3.6"
-  - "3.7"
```
daenney: Why are we removing 3.7 from the tests? 3.7 was release in June 2018 and is the current release of Python.

bastelfreak: in the current travis config, 3.7 is marked as `allow_failures` and the tests don't pass. I can readd that if you want. I had a quick look but wasn't able to fix the CI failures. I removed the job to make it run faster.

daenney: Mm, if the test don't pass lets leave that for now, but we should probably fix that before we go 1.0.

othalla: Yep, we should fix 3.7 for next release and add it back without allow_failures

othalla:
Looks fine

PR #508, opened 2019-05-30
Move to Python 3.6 for Docker

larsnaesbye:
Since Python 2.x [is getting deprecated as of January 2020](https://pythonclock.org/), shouldn't we move to 3.6 for the supplied Dockerfile?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:2.7-alpine
+FROM python:3.6-alpine
 
 ENV PUPPETBOARD_PORT 80
 EXPOSE 80


bastelfreak:
I'm always +1 for moving forward. Did you test the change?

larsnaesbye:
Well, no :-D 

But I saw that [others use 3.6 in their puppetboard Dockerfile](https://github.com/rsletten/puppetboard). I can test it out on Monday.

larsnaesbye:
Tested. It doesn't seem to work any different.

PR #511, opened 2019-06-15
Ignore facts environment for compatibility and performance

### LINKED ISSUE #496: Make showing facts work even if their environment doesn't match
Because of https://tickets.puppetlabs.com/browse/PUP-9290 , if you are using `puppet facts upload`, `facts_environment` of your nodes may not match `report_environment` and `catalog_environment`.

I propose to implement a workaround in puppetboard to get facts from `production` environment for nodes if they are not found in node's proper environment.

gdubicki:
Fixes #496

Because of https://tickets.puppetlabs.com/browse/PUP-9290 if someone
is using `puppet facts upload` their nodes have facts_environment set
always to "production". This breaks filtering in Puppetboard.
But it's both safe and beneficial to just ignore facts_environment
and depend only on catalog_environment to filter nodes by environment
because if a node is in env X then it mean that it has catalog_environment
set to X.

### DIFF
--- a/puppetboard/app.py
+++ b/puppetboard/app.py
@@ -131,7 +131,6 @@ def index(env):
     else:
         query = AndOperator()
         query.add(EqualsOperator('catalog_environment', env))
-        query.add(EqualsOperator('facts_environment', env))
 
         num_nodes_query = ExtractOperator()
         num_nodes_query.add_field(FunctionOperator('count'))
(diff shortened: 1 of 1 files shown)


gdubicki:
After reading https://tickets.puppetlabs.com/browse/PUP-9290?focusedCommentId=671276&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-671276 which shows how to set facts' environment when using `puppet facts upload` I find this PR not necessary anymore for compatibility, but it still makes sense for performance reasons.

bastelfreak:
I don't get how the coverage goes down if you delete code :thinking:

PR #512, opened 2019-06-15
Add feature for better performance in big Puppet envs

gdubicki:
Use new with_event_numbers params of `nodes` method in pypuppetdb
to not show exact numbers of changed resources in Overview and Nodes
lists to avoid using potentially slow event-counts query. Disabled
by default for compatibility with existing installations.

Requires https://github.com/voxpupuli/pypuppetdb/pull/153 to be merged, new pypuppetdb release and updated in requirements of puppetboard to work effectively, so I am WIPing it for now.

**UPDATE**: unWIPing as above MR has been merged. Thx @bastelfreak !


othalla:
@bastelfreak we should first check for https://github.com/voxpupuli/pypuppetdb/pull/153 as is it a requirement

bastelfreak:
https://github.com/voxpupuli/pypuppetdb/pull/153 got merged :)

PR #521, opened 2019-08-22
Revert "remove duplicate c3.js/d3.js links"

### LINKED ISSUE #519: Graphs do not load
Graphs are not loading for facts which have graphing enabled.  After clicking on the fact name the page simply says "processing" but never loads the graph.  Attached are some screenshots for reference.

Screenshot_2019-08-15_10-18-40
Screenshot_2019-08-15_10-16-20

blackknight36:
Fixes #519.  The c3.min.js and d3.min.js files are needed for graphs to work.


bastelfreak:
thanks for the fix @blackknight36 !

PR #530, opened 2019-10-31
Upgrade of tests requirements + resolving current deprecation warnings

dom-nie:
Upgrade to latest versions of packages for tests with all necessary changes.
Adding also ignore for DeprecationWarning, since third party packages like requests show some warnings, which can't be affected from this project perspective.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,0 +1,3 @@
+[pytest]
+filterwarnings =
+    ignore::DeprecationWarning
--- a/puppetboard/app.py
+++ b/puppetboard/app.py
@@ -858,7 +858,7 @@ def catalogs(env, compare):
     check_env(env, envs)
 
     if not app.config['ENABLE_CATALOG']:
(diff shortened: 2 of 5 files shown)


pytest.ini
```diff
@@ -0,0 +1,3 @@
+[pytest]
+filterwarnings =
+    ignore::DeprecationWarning
```
bastelfreak: please add the missing newline at the end of the file

bastelfreak:
@dom-nie thanks for the PR. Can you take a look at the failing travis jobs?

dom-nie:
@bastelfreak I went ahead and  opened #531 just to resolve this which I believe is best solution.
If you think we should keep Python2 let me know and I will upgrade this branch accordingly.

pytest.ini
dom-nie: Added.

bastelfreak:
let's rip it out in the other PR and afterwards rebase this one.

PR #531, opened 2019-10-31
Dropping support for python2.7 and python3.5 and adding python3.8

dom-nie:
Python 2.7 is deprecated in 2 months from now, so it's no longer necessary to support it for next release.
Python 3.8 is released, I think it's good to add it and drop Python 3.5.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,17 +1,14 @@
 dist: xenial
 language: python
 python:
-  - &latest_py2 2.7
-  - 3.5
   - 3.6
-  - &latest_py3 3.7
+  - 3.7
+  - &latest_py3 3.8
 
 jobs:
   fast_finish: true
   include:
   - python: *latest_py3
-    env: LANG=C
-  - python: *latest_py2
     env: LANG=C
   - stage: deploy (to PyPI for tagged commits)
     if: tag IS present
--- a/README.md
+++ b/README.md
@@ -13,10 +13,9 @@ As of version 0.1.0 and higher, Puppetboard **requires** PuppetDB 3. Version 0.3
 
 At the current time of writing, Puppetboard supports the following Python versions:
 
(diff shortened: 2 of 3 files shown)


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM python:3.7-alpine
+FROM python:3.8-alpine
```
bastelfreak: this currently triggers a warning:
```
[2019-10-31 19:53:55 +0000] [1] [INFO] Starting gunicorn 19.6.0
[2019-10-31 19:53:55 +0000] [1] [INFO] Listening at: http://0.0.0.0:80 (1)
[2019-10-31 19:53:55 +0000] [1] [INFO] Using worker: sync
/usr/local/lib/python3.8/os.py:1021: RuntimeWarning: line buffering (buffering=1) isn't supported in binary mode, the default buffer size will be used
  return io.open(fd, *args, **kwargs)
[2019-10-31 19:53:55 +0000] [7] [INFO] Booting worker with pid: 7
[2019-10-31 19:53:55 +0000] [8] [INFO] Booting worker with pid: 8
[2019-10-31 19:53:55 +0000] [9] [INFO] Booting worker with pid: 9
[2019-10-31 19:53:55 +0000] [10] [INFO] Booting worker with pid: 10
```

bastelfreak:
Thanks for the PR. Can you take a look at the inline comment I made?

Dockerfile
dom-nie: [Gunicorn issue](https://github.com/benoitc/gunicorn/issues/2091#issuecomment-547786011)
I guess it's best for now to keep Python37 in dockerfile and upgrade after they will release 20.0.0

dom-nie:
@bastelfreak  Ok, I have changed dockerfile to Python3.7 as gunicorn below 20.0.0 won't work with Python3.8 and it's  still unreleased.
I think best course of action would be to upgrade dockerfile in seprate PR after they will release version with fix.

PR #532, opened 2019-11-01
Adding mypy + Cleanup + CommonMark upgrade to 0.9.1

dom-nie:
Cleanup of unused code, getting rid of star import from pypuppetdb and replacing imported libraries with local imports.
Added mypy to tests.
CommonMark bumped to 0.9.1, since #497  was merged.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@
 *.egg-info
 dist
 build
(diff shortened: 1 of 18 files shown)


bastelfreak:
Thanks for the PR!

PR #538, opened 2019-12-18
Cast inventory data toString.

johnzimm:
This prevents a type error for fact values in the inventory page that are numeric.  Some of our custom facts simply return a numeric count which causes the replace function to fail unless it's cast as a string.


bastelfreak:
thanks for the fix!

PR #545, opened 2020-01-27
Update app.py

pjonesIDBS:
urllib.parse is for Python 3.x and doesn't work in 2.x. This totally broke the Puppetboard. The above will fix the issue for this file. The same problem needs fixing in api.py from https://github.com/voxpupuli/pypuppetdb/blob/master/pypuppetdb/api.py

Unfortunately, it looks like Python 2.7 support was intentionally removed from PyPuppetDB. This isn't helpful when it essentially totally breaks this module :(

### DIFF
--- a/puppetboard/app.py
+++ b/puppetboard/app.py
@@ -1,8 +1,11 @@
 from __future__ import absolute_import
 from __future__ import unicode_literals
 
-
-from urllib.parse import unquote, unquote_plus, quote_plus
+try:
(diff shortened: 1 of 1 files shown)


pjonesIDBS:
I've created a new PR instead making the default Python 3, with fallback to Python 2. Hopefully that will pass the build tests.

pjonesIDBS:
https://github.com/voxpupuli/puppetboard/pull/548

PR #547, opened 2020-01-28
Fixing the requirements.txt as it's breaking Puppetboard.

pjonesIDBS:
This module is NOT compatible with Python 3.x The latest versions of PyPuppetDB (>2.x) are Python 3 only. 

Until PuppetBoard is updated to use Python 3 please change this back to ensure the module works. This change was obviously not correctly tested as it breaks Puppet board.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,6 +5,6 @@ MarkupSafe >=1.1.1
 WTForms >=2.2.1
 Werkzeug >=0.16.0
 itsdangerous >=1.1.0
-pypuppetdb >=2.0.0
+pypuppetdb <2.0.0
 requests >=2.22.0
 CommonMark==0.9.1


bastelfreak:
Thanks for the fix. Are you also able to provide a patch so it works with python 3? Or update the tests so they fail properly on python 3?

pjonesIDBS:
Hi @bastelfreak, I think the whole module will need to be revisted to make sure it's Python 3.x compatible. This is quite a big piece of work and at this time I don't have a suitable dev environment available to work with. If I do start looking at this I'll need to do so on my own time.

pjonesIDBS:
This pull request is still required to get things working again: https://github.com/voxpupuli/puppetboard/pull/545

An alternative might be to reverse the try and except so that Python 3 is the default, and Python 2 the fallback. That way it might pass the automated tests. The main thing is it should start working again for existing users.

bastelfreak:
@pjonesIDBS could you make python 3 the default? If the tests are green I can merge it.

pjonesIDBS:
Hmm, I've done that https://github.com/voxpupuli/puppetboard/pull/548 but sadly the build tests still fail :(

PR #548, opened 2020-01-31
Update app.py to support fallback to Python 2.7

pjonesIDBS:
Currently Puppetboard doesn't work on Systems still using Python 2.7 wsgi so we need to try and fallback on the old urllib import.

### DIFF
--- a/puppetboard/app.py
+++ b/puppetboard/app.py
@@ -1,14 +1,12 @@
 from __future__ import absolute_import
 from __future__ import unicode_literals
 
-
-from urllib.parse import unquote, unquote_plus, quote_plus
+# load python 3, fallback to python 2 if it fails
 try:
-    from urllib.parse import unquote, unquote_plus, quote_plus  # Python 3+
+    from urllib.parse import unquote, unquote_plus, quote_plus
 except ImportError:
-    # Fallback for Python 2.x support
-    from urllib import unquote, unquote_plus, quote_plus  # Python 2.X
-    
+    from urllib import unquote, unquote_plus, quote_plus  # type: ignore
+
 from datetime import datetime, timedelta
 from itertools import tee
 import sys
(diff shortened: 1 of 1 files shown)


bastelfreak:
@pjonesIDBS can you test if this works for you?

mattock:
For what it's worth the fallback seems to work ok. Right now my setup is
* ubuntu/bionic64 vagrant box, version 20200206.0.0
* puppet-puppetboard 6.0.0
* puppetboard commit c215ebd0a734b2 (latest from "master" right now)
* Latest Puppet 6 version (puppetserver, puppetdb, puppet-agent)

Without this patch I get
```
Traceback (most recent call last):
  File "/srv/puppetboard/puppetboard/wsgi.py", line 15, in <module>
    from puppetboard.app import app as application
  File "/srv/puppetboard/puppetboard/puppetboard/app.py", line 5, in <module>
    from urllib.parse import unquote, unquote_plus, quote_plus
ImportError: No module named parse
```
With this PR applied I get:
```
Traceback (most recent call last):
  File "/srv/puppetboard/puppetboard/wsgi.py", line 15, in <module>
    from puppetboard.app import app as application
  File "/srv/puppetboard/puppetboard/puppetboard/app.py", line 13, in <module>
    from flask import (
ImportError: No module named flask
```
The latter is the same error I got when testing an old version of Puppetboard (e.g. 2c3a593c5d27) which used to work with my old Puppet 5-based setup on Ubuntu 18.04.

mattock:
I was able to make Puppetboard work with this patch. The problem was that I had used puppetlabs-apache to purge configs:
```
  class { '::apache':
    purge_configs => true,
    --- snip ---
```
When I set ```purge_configs => false```, ran ```a2dismod mpm_event`` _and_ applied this patch Puppetboard finally started working.

pjonesIDBS:
@bastelfreak Sorry, I've not been able to reply before... I can confirm the change works. At the moment, we're using a forked version of the repo with the patch applied :)

nmaludy:
Implemented in https://github.com/voxpupuli/puppetboard/pull/555

Issue #529, opened 2019-10-04
jquery out of date


anarcat:
puppetboard ships with jquery 2.1.1 which is has been released over 4 years ago. it's also not the latest supported version of the 2.1.x series (2.1.4). and obviously it's not the latest jquery release either (now at 3.4.1).

this makes it hard to package in debian (among other thigns) because Debian (currently) ships with jquery 3.3. I also suspects it might have security vulnerabilities, at least cvedetails lists two XSS:

https://www.cvedetails.com/vulnerability-list/vendor_id-6538/product_id-11031/version_id-286384/Jquery-Jquery-2.1.1.html

so it seems like upgrading jquery might be a good idea here. :p

larsnaesbye:
Seconded. Any arguments against jumping to 3.4.1 right away?

bastelfreak:
I'm always +1 for up2date dependencies. Anybody here wants to propose and test a PR?

kevgreen-ebay-com:
With thanks to the supportive open source contribution friendly policy of my employer, I've got some patched code for this.

Long story why, but I couldn't properly test a patch against the latest version. so I've got a patched version of the 1.1.0 tree with upgraded jquery ready to upload to a branch here, but no access. If someone could grant me access to create a new branch for it, or suggest an alternative way for me to upload I'll do it ASAP.

It looks like nothing changed regards jquery between 1.1.0 and latest version anyway, so this should just apply cleanly to the latest.

PS. It would be great if python 3.5 support for the current latest could be put back and kept a little while longer.

anarcat:
to provide patches here, you need to "fork" the repo, push patches to your git repo, and issue a "pull request". more details on how pull requests work is available here:

https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/about-pull-requests

if i misunderstood your question: sorry for stating the obvious. :)

also feel free to just copy-paste the patch here if that process is too cumbersome - it's better to have a half-working suggestion than nothing at all!

same comment regarding py3.5 - maybe we could still support that if the patch is minimal, but that should be a separate issue.

kevgreen-ebay-com:
Ah, yes, that's a pattern I'm familiar with, just not really done it that way before. Hopefully I've done it right as above :)

https://github.com/voxpupuli/puppetboard/pull/549/files

anarcat:
thanks for the patch! and yes, you did fine :)

PR #549, opened 2020-02-10
bump jquery to 3.4.1 and update linking

kevgreen-ebay-com:
https://github.com/voxpupuli/puppetboard/issues/529


kevgreen-ebay-com:
I've tested this doesn't generate any HTTP errors, except favicon.ico, but I don't think that's provided in the code anyway, and had a glance at my browser's script console and don't think it's caused any extra troubles. There is a possibility of incompatibility between jquery and related js code that sits alongside it, but I would not know how to test for that really.

kevgreen-ebay-com:
Honestly, I really haven't put anything nefarious on top of the distribution version of jquery:

$:puppetboard/puppetboard/static/jquery-3.4.1$ md5sum jquery.min.*                                                        220afd743d9e9643852e31a135a9f3ae  jquery.min.js
04b42631197bc74a486afe5b76699779  jquery.min.map
$:puppetboard/puppetboard/static/jquery-3.4.1$ wget -q https://code.jquery.com/jquery-3.4.1.min.map https://code.jquery.com/jquery-3.4.1.min.js
$:puppetboard/puppetboard/static/jquery-3.4.1$  md5sum jquery-3.4.1.min.*
220afd743d9e9643852e31a135a9f3ae  jquery-3.4.1.min.js
04b42631197bc74a486afe5b76699779  jquery-3.4.1.min.map
$:puppetboard/puppetboard/static/jquery-3.4.1$ 

Now, whether or not you trust the vendors of jquery itself not to do anything nefarious, well, that is another matter ;)

anarcat:
> Honestly, I really haven't put anything nefarious on top of the distribution version of jquery:

well of course you'd say *that*. :) that's not the point, the point is for us to figure out how we make sure you're telling the truth, or maybe avoid that requirement altogether so that we don't have that awkward conversation again in the future. :p

> Fetching it as a part of the build process probably makes a lot more sense, but you'd need a sane workaround for the versioned path and filename situation in your pipeline, while still being able to demonstrate which version of jquery belongs with which version of puppetboard. Shouldn't be too hard though.

maybe we could commit checksums in the build system, for example. alternatively what we do in Debian packaging is we check upstream OpenPGP signatures so that we know we at least get what we expect from upstream.

now whether we actually trust upstream is a whole other ball game of course, but maybe we can assume enough eyeballs are looking at jquery for that not to matter for this specific project?

PR #555, opened 2020-03-19
Fixed bugs in metrics v2 usage

nmaludy:
Tried using the latest version of `puppetboard` against the recently released PuppetDB 6.9.1 and ran into several problems related to them disabling the metrics/v1 endpoint (https://puppet.com/security/cve/CVE-2020-7943/) along with the bug fixes for metrics/v2 that were recently merged.

**Fixes**
- In Metrics v2 the `avg-resources-per-node` metric now returns a string in the format `1234/11` (`<total_resources>/<num_nodes>`). This displayed funny on the UI and was not consistent with previous behavior. To fix this i reverted back to computing the average resources per node as we were doing with metrics v1.
- In Metrics v2 the way you list all available metrics as done in the Metrics page changed. This page failed to load on the latest version. To support these changes i needed to update `pypuppetdb` (PR: https://github.com/voxpupuli/pypuppetdb/pull/175). We also had to do some data manipulation on the back end because the formatting of the response changed in PuppetDB.

**TODO**
- [x] Merge in pypuppetdb pr: https://github.com/voxpupuli/pypuppetdb/pull/175
- [x] Updated `requirements.txt` with new pypuppetdb version

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ Development
 
 * Fixed Puppetboard's usage for the new metrics v2 API both on the home page for computing the average resources/node and the `Metrics` listing page. This change now supports the changes in PuppetDB >= 6.9.1 (https://puppet.com/security/cve/CVE-2020-7943/)
 * pypuppetdb: raise version requirement `>=2.1.0` because changes were needed in this library to support the metrics v2 fixes.
+* app.py: Added python2 backwards compatability fix for importing `urllib`.
 
 2.0.0
 ----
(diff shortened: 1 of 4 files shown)


bastelfreak:
Looks good to me. I will try to ship a new pypuppetdb release and afterwards this PR can be rebased.

nmaludy:
@bastelfreak once we get https://github.com/voxpupuli/pypuppetdb/pull/175 merged and released, just need to update to latest version in `requirements.txt` then this is good.

sigv:
Does this intentionally not cover Puppet Server 5.3?

nmaludy:
@sigv what do you mean?

sigv:
Puppet Server [5.3.12](https://puppet.com/docs/puppetserver/5.3/release_notes.html#puppet-server-5312) was released with the same breaking change of disabling v1 metrics endpoints as the 5.x branch is still being patched. As Puppet 5 is still having these kinds of patches, I was thinking about those who have not switched to Puppet 6 for any reason, and how this PR does not cover them.

nmaludy:
Would you be willing to submit a patch that covers them? Otherwise they can upgrade to 6.

sigv:
We have upgraded to 6.x now so we are fine, however as 5.x is not EOL I believe it makes sense to cover it. Opened a PR for it. :+1:

PR #557, opened 2020-03-21
fix path to daily_reports_chart.json

vchepkov:
provide relative path to daily_reports_chart.json on 'Overview' page
and absolute path on node page, since node page url is dynamic

Currently charts are broken, in case dashboard is not located in the root of the web site


vchepkov:
I honestly don't know how

nmaludy:
OK I'll see if I can lend a hand

jplindquist:
Thanks for this, works great!

Issue #553, opened 2020-03-11
Puppetboard v2.0.0 incompatible with latest Puppet Server 6.x and 5.x


sigv:
Puppetboard (as of v2.0.0) uses [v1 metrics](https://puppet.com/docs/puppetserver/6.3/metrics-api/v1/metrics_api.html) instead of [v2 (Jolokia) metrics](https://puppet.com/docs/puppetserver/6.3/metrics-api/v2/metrics_api.html). The v1 metrics API has been deprecated for a while and as of Puppet Server releases [6.9.1](https://puppet.com/docs/puppetserver/latest/release_notes.html#puppet-server-691) and [5.3.12](https://puppet.com/docs/puppetserver/5.3/release_notes.html#puppet-server-5312) (10 March 2020) these endpoints are disabled by default due to [CVE-2020-7943](https://puppet.com/security/cve/CVE-2020-7943). Trying to load pages from Puppetboard results in:

```text
ERROR:pypuppetdb.api: localhost:8080 over HTTP.
ERROR:puppetboard.utils:404 Client Error: Not Found for url: http://localhost:8080/metrics/v1/mbeans/puppetlabs.puppetdb.population%3Aname%3Dnum-nodes
```

jvperrin:
From all I can tell from reading through the code this is a pypuppetdb issue (one of puppetboard's dependencies) instead of a puppetboard issue. When fetching metrics, [puppetboard is using `puppetdb.metric`](https://github.com/voxpupuli/puppetboard/blob/b67331a3e419d3c998beeae1552d731301bf9c6e/puppetboard/app.py#L116-L125) which is calling out to [this function in `pypuppetdb`](https://github.com/voxpupuli/pypuppetdb/blob/9eb9fb64c016bcee7c4cb52f458254afc84c9a9a/pypuppetdb/api.py#L887-L895) which uses [this v1 metrics mbean endpoint](https://github.com/voxpupuli/pypuppetdb/blob/9eb9fb64c016bcee7c4cb52f458254afc84c9a9a/pypuppetdb/api.py#L26), so I think this needs to be fixed on the pypuppetdb side of things and then the updated release there pulled into in this repo.

sigv:
Thank you for the explanation! Just opened voxpupuli/pypuppetdb#172 for this.

sigv:
This actually raises a further issue about Python 3 support, which could be considered #551.

[...]

PR #558, opened 2020-03-22
Adjust metrics API for Puppet Server 5.3.x

sigv:
Puppet Server releases done on 10 March 2020 forcefully disabled the metrics v1 endpoints. The appropriate change was made for Puppet Server 6.9.1 support, however Puppet Server 5.3.12 also needs the same treatment.

Although everyone is urged to upgrade when possible, in case someone wants to use Puppetboard and has not upgraded yet, we should support them, as the ['Puppet packages and versions'](https://puppet.com/docs/puppet/latest/about_agent.html#puppet-releases-and-lifecycle) docs page currently states that Puppet 5.5.x (Puppet Server 5.3.x) is expected to become EOL in November 2020.

Ref: #553, #555

### DIFF
--- a/puppetboard/app.py
+++ b/puppetboard/app.py
@@ -84,8 +84,10 @@ def check_env(env, envs):
 def metric_params(db_version):
     query_type = ''
 
-    # PuppetDB moved to a new metrics API (v2) in 6.9.1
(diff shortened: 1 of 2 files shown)


bastelfreak:
thanks for the patch!

PR #563, opened 2020-04-07
Puppet DB 5.2.13 requires v2 metrics

rtkennedy:
Puppet is now forcing the v2 metrics endpoint in PuppetDB 5.2.13.


bastelfreak:
thanks for the fix!

PR #575, opened 2020-06-12
Use a select for endpoint selection in query

raphink:
Use a select field to choose API endpoint in query, default to PQL.
image


raphink:
I have no idea why the tests are not passing…

PR #587, opened 2020-07-31
Improve README badges

raphink:
image

Or do you prefer another background color for the Vox badge?

#8ac9ff: 8ac9ff
#3f51b5: 3f51b5

Or even split:

Split


raphink:
@bastelfreak so you prefer the default (orange background) vox pupuli badge?

PR #591, opened 2020-09-21
Fix fact sorting for uptime

jgrammen-agilitypr:
an attempted fix that doesnt appear to work, and may have consequences for other pages


gdubicki:
Hi @jgrammen-agilitypr ! I think that you almost have it working.

I would just add an `if` to apply the code that you have added here but only if the fact name is "uptime".

To do that you could add an extra boolean parameter to the `datatable_init` macro, like a "natural_time_delta_sort", false by default.

And the value of this flag could be passed from the `def fact()`.

jgrammen-agilitypr:
ok, I have rebased on voxpupuli:master
I have added the script to the static folder for offline mode.
the `datatable_init` flag has been created and is being used in the template
but I could not figure out the change tot he `def fact()` function

gdubicki:
Thanks @jgrammen-agilitypr !

I meant that I imagine that we want to enable the natural time delta sorting only for the specific facts, like "uptime".

So in the `def fact()` (see the code [here](http://github.com/voxpupuli/puppetboard/blob/to_csv/puppetboard/app.py#L683-L716)) we should pass an extra parameter like `natural_time_delta_sort` to the `'fact.html'` template. The value of `natural_time_delta_sort ` should be boolean saying the the `fact` is one of those that are natural-time-delta-sortable.

And then we can pass this parameter as a value for `natural_time_delta_sort` in the `datatable_init` macro call [here](http://github.com/voxpupuli/puppetboard/blob/to_csv/puppetboard/templates/fact.html#L9-L9).

jgrammen-agilitypr:
ok, I made more adjustments, moved the natural_time_delta_sort around. and made an attempt at modifying the fact function.

puppetboard/app.py
```diff
@@ -708,15 +708,17 @@ def fact(env, fact, value):
             value_json = value_object
         else:
             value_json = dumps(value_object)
-
+    natural_time_delta_sort = True
     return render_template(
```
gdubicki, line 712: This should be set to True only if the name of the fact is one of „uptime”, etc.

puppetboard/templates/_macros.html
```diff
@@ -79,6 +79,9 @@
     {% else -%}
     {
       "targets": -1,
+      {% if natural_time_delta_sort %}
```
gdubicki, line 82: I think that this should go above, under the case where `table_html_id == 'facts_table'` as we are changing the sort in that table.

gdubicki:
Almost there @jgrammen-agilitypr, check out my comments please.

jgrammen-agilitypr:
I made my best attempt at correcting your 2 comments. Thank you for your patience with me as I struggle through this.

gdubicki:
I made it work with my commit, @jgrammen-agilitypr. Please check out its commit message to know what I changed. :)

Please just rebase and in my opinion then it will be ready to merge. :) 

(We also think that in the long term we should rather switch to DataTables plugins to:
* detect the type of our values, f.e. [IP address](https://datatables.net/plug-ins/type-detection/ip-address),
* sort the columns based on the detected type, f.e. [IP addresses numerically](https://datatables.net/plug-ins/sorting/ip-address).
Then we won't have to have almost anything in the Python code. I am just worried that DataTables type detection & sorting plugins don't go in hand with custom renders - see that we had to read the raw value to sort with from the "title" field.)

jgrammen-agilitypr:
My branch has been rebased on the 'voxpupuli:master' branch. If you would prefer that I abandon this work and and make some attempt at datatables orthogonal data, just let me know

gdubicki:
I prefer to be pragmatic, so in my opinion if we have a working feature here and the code is ok, then we should merge, release it and let ourselves and the community already use what you did.

If you have the time and will to start working on a more mature and universal solution then please go ahead and to it in a separate PR.

But I am not the only (de facto) maintainer here - what are your opinions, @bastelfreak / @mterzo / @corey-hammerton ?

bastelfreak (approved):

PR #594, opened 2020-10-01
Convert Boolean Parameters to boolean for puppetdb query

### LINKED ISSUE #583: When viewing a specific fact with a True or False value, no nodes will display in the resulting list
Example: /fact/is_virtual/False
Viewing this URI will return an empty list of nodes despite many of my nodes having this fact set as False. I am able to repeat the issue on any fact with a True or False

sebastianrakel:
A PuppetDB Query with a Boolean Parameter must not be in quotes. This PR fix this issue and resolves #583

### DIFF
--- a/test/test_utils.py
+++ b/test/test_utils.py
@@ -217,3 +217,30 @@ def test_stop_http_error():
     gen = utils.yield_or_stop(my_generator())
     for val in gen:
         assert 1 == val
+
+
+def test_is_bool_false():
+    test_values = [
+        "Test",
+        "SomeWrongData",
+        2,
+        2.0,
+        -5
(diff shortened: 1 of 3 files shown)


test/test_utils.py
```diff
@@ -217,3 +217,32 @@ def my_generator():
...
+        "False",
+        "false",
+        "1",
+        "0",
+        1,
+        0
+    ]
+
+    for test_value in test_values:
+        assert utils.is_bool(test_value)
```
bastelfreak, line 248: please add the missing `\n` at the end of the line

bastelfreak (approved):

Issue #572, opened 2020-05-13
on the node detail page, facts are ordered in reverse alphabetical order


epackorigan:
When looking at the details of a report for a given node, the facts on the right side are ordered in reverse alphabetical order. Would it be possible to get the order in ascending order for that table?

gdubicki:
Hi @epackorigan !

Thanks to @jokreliable 's contribution this will be fixed in the next puppetboard release.

gdubicki:
Fixed in v3.0.0.

PR #597, opened 2020-11-03
changing the default sort order for facts_table

jokreliable:
This would address #572

### DIFF
--- a/puppetboard/templates/_macros.html
+++ b/puppetboard/templates/_macros.html
@@ -55,8 +55,12 @@
     "pageLength": {{ default_length }},
     // Search as regex (does not apply if serverSide)
     "search": {"regex": true},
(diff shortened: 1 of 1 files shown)


jokreliable:
I don't quite understand why the ci is failing on various versions of python with my change...

gdubicki:
Thanks for the contribution @jokreliable !

This change looks very reasonable, I cannot find any trace of information why is the order `desc` instead of `asc`...

The only slight doubt about this PR that I have is that perhaps we should just change the order to `asc` everywhere, without the `if`.

But I'll just merge it for now to not delay this further. :)

jokreliable:
Regarding the all asc, that would be good, I think, except for reports (where it's likely better to see the latest one on top). That being said, I'm not that familiar with the entire code base, so I could be wrong here...

In any case, thanks for merging that in!

PR #603, opened 2021-01-23
Update README.md

mathse:
GRAPH_FACTS do not work if you put your facts into single quotes - so this is just a correction on the documentation


kenyon:
There must be something else going on for you. It's just normal shell command line, so it's equivalent with or without quotation marks (for simple values that don't need quoting):

```console
kenyon@einstein ~ % docker container run --rm --env VAR1='abc,def' fedora env | grep VAR1
VAR1=abc,def
kenyon@einstein ~ % docker container run --rm --env VAR1=abc,def fedora env | grep VAR1
VAR1=abc,def
```

PR #604, opened 2021-01-24
Migrate from Travis to GitHub Actions and add upper bounds in requirements.txt

gdubicki:
with testing, publishing to PyPI and creating a release in GitHub.

Note that notifications to IRC have NOT been migrated yet. Let's do that in a separate PR. (And maybe someone who is using IRC unlike me would be better to do it...?)

Notes:
* To make publishing to PyPI work please set `PYPI_API_TOKEN` secret in the GitHub repo config,
* Uncomment line 29 in `.github/workflows/publish-to-pypi.yml` to test releasing to PyPI (note: I have not tested if this actually works - in my project I just made a few pre-releases).


gdubicki:
You can see how the GitHub Actions outputs will look like here: https://github.com/gdubicki/puppetboard/actions

PR #605, opened 2021-01-26
Fix auto-resize in radiator view

chrisongthb:
This fixes the links to jquery 3.5.1
See also https://github.com/voxpupuli/puppetboard/pull/592 and https://github.com/voxpupuli/puppetboard/commit/558597c74d16e4cde32c8dcb5b01dee9fe2b1c36


bastelfreak:
thanks for the fix!

PR #606, opened 2021-01-26
Manage any other exception for get_or_abort

qkponton:
Log any other exception that can be raised from execution of a function and raised 500


bastelfreak:
thanks for the PR!

PR #610, opened 2021-02-12
Get the info about test results and coverage in PRs

gdubicki:


bastelfreak:
coveralls trolls us again. that's a false positive and IMO safe to merge.

Issue #544, opened 2020-01-23
Front Page - can we customize the whole thing?


tskirvin:
I've got a puppetdb that supports ~2500 hosts.  The front page "resources managed" and such queries tend to bog down the overall page load time on first load of the day enough that I can *expect* to have the request time out, even after I've disabled the reports-per-day graph and such.

Could you just provide a way to customize the front page in most obvious details?  Decide whether to get total resource counts; the graphs; the list of nodes with problems; and so forth.

Issue #533, opened 2019-11-05
Remove nodes status details from overview page


fangludi:
Is there a way to remove "Nodes Status Details" from overview page; sometimes if thousands of machines (such as during outage window) shut down, the overview page will take a long time to open because there're thousands of unreported machine listed in the Nodes Status Details table.

PR #614, opened 2021-03-19
Configurable front page

gdubicki:
Implements #544 and #533.

Also organizes the docs for the settings in README.


gdubicki:
A screenshot showing the updated front page with both sections disabled. Note where the "total population" metric is located now.
Screenshot 2021-03-19 at 22 21 27

gdubicki:
Hm, I need to fix the tests...

gdubicki:
I also forgot to update `def index(env)`, so we still actually get the data that we do not show in the template... 🤦

gdubicki:
Oups, disabling `NODES_STATUS_DETAIL_ENABLED` breaks failed/pending/changed/unreported metrics...

sebastianrakel (approved):
This is fine, but the decreasing test coverage should be fixed

gdubicki:
Thanks @sebastianrakel , but I have to make it work with RESOURCES_STATS_ENABLED=True and NODES_STATUS_DETAIL_ENABLED=False without actually all the processing that would be done if NODES_STATUS_DETAIL_ENABLED=True.

sebastianrakel:
After a chat with @gdubicki, we will close this pull request in favor of #639

PR #618, opened 2021-04-07
Improve facts columns balancing

smortex:
Grouping facts by first letter is inefficient when a lot of facts begin
with some specific letters.  E.g. with multiple network interfaces.

Before:
Before

After:
Screenshot_2021-04-06 Puppetboard(1After)


gdubicki:
This looks great, thanks for you contribution @smortex !

PR #629, opened 2021-08-04
Enumerate FreeBSD package

smortex:
PupeptBoard is available as a package on FreeBSD.  Support for managing
PupeptBoard using system packages is WIP in puppet-puppetboard.


gdubicki:
Thank you for your contributions, @smortex ! Most of all for maintaining the FreeBSD package. 😊

PR #630, opened 2021-08-04
Fix CI

smortex:
Some dependencies are missing.

I could get the list with:

```
mypy --install-types
```

### DIFF
--- a/requirements-test.txt
+++ b/requirements-test.txt
@@ -13,3 +13,6 @@ bandit
 coveralls
 mypy
 tox
+types-requests
+types-setuptools
+types-toml


kenyon (approved):

gdubicki:
This is cool, but maybe adding the `mypy --install-types` command to the CI script (and contributing docs, probably) would be even better, @smortex ?

smortex:
> This is cool, but maybe adding the `mypy --install-types` command to the CI script (and contributing docs, probably) would be even better, @smortex ?

It was my first move: I added this to the "Install dependencies" step, but when you do this you have to tell it what source to consider (see [this CI failure](https://github.com/voxpupuli/puppetboard/runs/3237245219?check_suite_focus=true): automatic installation works when the tests failed previously, if the test where never run you have to pass the files you want to check). So I tried with `mypy --install-types --non-interactive .` [as suggested in this part of the documentation](https://mypy.readthedocs.io/en/latest/running_mypy.html#library-stubs-not-installed): it installs proper dependencies but it also run the checks.

So this was running a part of the tests twice, which might not be great…  I then considered explicitly listing the dependencies as above…

That being said, I am not used to Python at all and maybe we can automatically discover dependencies without running the tests.  However, I failed to find a way in the doc.

gdubicki:
Got it, @smortex . Thanks!

Issue #625, opened 2021-06-01
Feature Request - Add datetime filter to Reports


mike406:
It would be very useful to be able to filter the Reports history by a date and time. It's not very practical to have to click through the pagination to get to a certain time in the past, and in some cases is impossible depending how far back into the history you want to go.

sebastianrakel:
@mike406 i provided a PR with a screenshot in it, this should fix your issue

gdubicki:
Released in v3.2.0. :)

PR #638, opened 2021-10-04
Add from/till filter for reports

sebastianrakel:
this should fix issue #625


sebastianrakel:
The default behavior is both fields are empty

image

gdubicki:
LGTM, thanks for the contribution @sebastianrakel !

PR #643, opened 2021-11-20
Fix getting nodes with non-string fact values

### LINKED ISSUE #612: Numerical facts don't get filtered properly
I noticed an issue with trying to filter numeric facts. For example I need to track version upgrades for our 600 Windows 10 PCs so I use the /fact/windows_release_id fact. If I want to view version 1909 for example, I would put /fact/windows_release_id/1909. However when I do this, it shows me all Windows 10's regardless of version number. I instead have to use the search box and type in "1909" to get the filter I need.

gdubicki:
Examples:
* int (in this case the filtering did not occur at all
and ALL results were returned),
* list of dict (here the filtering was not done and
ZERO results were returned),

Fixes #612.

### DIFF
--- a/puppetboard/app.py
+++ b/puppetboard/app.py
@@ -4,7 +4,6 @@ from __future__ import unicode_literals
 from urllib.parse import unquote, unquote_plus, quote_plus
 from datetime import datetime, timedelta
 from itertools import tee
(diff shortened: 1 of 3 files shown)


gdubicki:

Issue #567, opened 2020-04-14
Missing trailing 'u' character in facts with hashes


yakatz:
In an educational institution, we have a very large number of facts that end with `.edu`, but in puppetboard, all of them inside hashes show up as `.ed `. Running `puppet facts` on the systems show the correct value.

For example, `fqdn` is correct, but `trusted['certname']` is missing the trailing `u`.

Puppetboard:
![image](https://user-images.githubusercontent.com/93333/79252763-1dc2a380-7e50-11ea-87b2-0c22c126fe02.png)

And in Chrome inspector:
![image](https://user-images.githubusercontent.com/93333/79253050-97f32800-7e50-11ea-9557-bd5f4b4fd360.png)

epackorigan:
I see the same where `os::name` shows `Ubunt` for a lot of my machines. Same for `os::id`. `operatingsystem` is showing the correct value.

epackorigan:
It appears that there is code removing the u at the end of values [here](https://github.com/voxpupuli/puppetboard/blob/c72d3e88639a5c62c9a3577c5efcc9c464552fbf/puppetboard/templates/_macros.html#L65)
I don't have enough context to understand why that code is doing this there.

yakatz:
Starting a string with `u'` in python means it is Unicode-encoded. I am not sure how I missed this in my searching. I will see if I can work up a patch.

gdubicki:
Fixed in v3.3.0 released today. Please check it out @yakatz . :)

yakatz:
Thanks. Resolves the issue.

PR #644, opened 2021-11-21
Format structured fact values as pretty-printed, colored JSON

gdubicki:
and extend the width of the fact values column in the node
view as the values are usually longer that the names.

The colors mean:
* orange - number,
* green - dict's key name,
* blue - boolean,
* magenta - null (not sure if this will be used at all as null facts are not pushed to puppetdb, are they?),

Note: the downside of using JSON in JS code is that float numbers with 0 decimal numbers, f.e. `2.0`, are shown like they are ints, f.e.`2`.

### DIFF
--- a/puppetboard/app.py
+++ b/puppetboard/app.py
@@ -12,6 +12,7 @@ from flask import (
 
 import logging
 import json
+from json import dumps
 
 from pypuppetdb.QueryBuilder import (ExtractOperator, AndOperator,
                                      EqualsOperator, FunctionOperator,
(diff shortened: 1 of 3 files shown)


gdubicki:
Before:

Screenshot 2021-11-21 at 13 18 40

After:

Screenshot 2021-11-21 at 13 20 14

gdubicki:
This would fix #567 as this PR gets rid of the old code that removes the trailing `u`.

gdubicki:
It was not meant as it, but you could think of this as an implementation of  #83.

bastelfreak (approved):
I think using standard json is a good idea

gdubicki:

PR #653, opened 2021-12-28
Add friendly message for invalid query and 0 results

gdubicki:
Add friendly message for invalid query and 0 results
    
    instead of, respectively, showing the raw HTTP 400 without
    the specific reason and not showing anything.
    
    Also:
    * show the number of results when they are > 0,
    * use fixed-width font to query textarea and the error
      message (if there is one), to point to error in the
      the query precisely,
    * remove the nonsense "cancel" button, make the "submit"
      look active,


gdubicki:
This is how it looks like:
* zero results:
Screenshot 2021-12-28 at 19 09 04

* error in query:
Screenshot 2021-12-28 at 18 48 19

* normal result:
Screenshot_2021-12-28_at_18_48_50

gdubicki:
Originally I wanted to finish #429 and this was supposed to be just a warmup, but it turned out to be enough work to release it as a separate PR.

bastelfreak (approved):

Issue #428, opened 2017-10-19
Export buttons for tables


emlun007:
I am trying to create an export button that can copy and export the tables as csv, excel and pdf files.

I have loaded following libraries:
https://cdn.datatables.net/buttons/1.4.2/js/dataTables.buttons.min.js 
//cdn.datatables.net/buttons/1.4.2/js/buttons.html5.min.js 

and placed following code in layout.html:

    {% block script %} {% endblock script %}
    <script type="text/javascript">
    {% block javascript %} {% endblock javascript %}
      $(document).ready(function(){
          $(".ui.dropdown").dropdown();
          $.getScript('{{url_for('static', filename='js/lists.js')}}')
          $.getScript('{{url_for('static', filename='js/tables.js')}}')
          {% block onload_script %} {% endblock onload_script %}
        })

    
      $("table:visible").DataTable( {
          dom: 'Bfrtip',
                buttons: [
      
                  'copy',
      
                  'csv',
      
                  'excel',

	          'pdf',

                  'print',
      
              ]
     
    
      } );
    </script>
    {% block head %} {% endblock head %}

I cannot see the export buttons anywhere. What is the wrong in my code ?

mterzo:
Post the code as a work in progress merge request. I see lots of things here but hard to give context with out the diff.

larsnaesbye:
It's in #429 apparently :)

emlun007:
Update: button in #429 works now on Inventory and Catalogs

emlun007:
Update: button in #429 works for all pages. The issue with broken titles is fixed

gdubicki:
Partially implemented - for the Query results - and released in [v3.4.0rc2](https://github.com/voxpupuli/puppetboard/releases/tag/v3.4.0rc2).

PR #654, opened 2021-12-29
Add download as CSV/XLSX to Query results + update static assets

gdubicki:
Additionally update jQuery and other JS & CSS related to DataTables.


gdubicki:
This is how the buttons and query results in non-JSON mode will look:

Screenshot 2022-01-15 at 15 31 32

gdubicki:
This partially implements #428. Extending it to cover all the tables can be added in another PR and I would be happy if someone else would like to do it. :)

bastelfreak (changes requested):
updating the static files doe not relate to the new feature? please put that into a dedicated PR. Do you have any semiautomatic way for updating the files? If so, please document that.

gdubicki:
> updating the static files doe not relate to the new feature? please put that into a dedicated PR. Do you have any semiautomatic way for updating the files? If so, please document that.

That request makes a lot of sense. However as I was adding a new feature that is using a new library I have added it in the latest stable version. And to avoid compatibility issues I updated the related libraries too.

Theoretically I could downgrade Datatables Buttons to a version compatible with the version of Datatables that we had before the upgrade but I don't even know which that is - they are versioned separately and [the website doesn't allow choosing older versions on their download page](https://datatables.net/download/index).

Because of that and because we didn't upgrade any _major_ version here, I think that we can keep this change in this PR.

bastelfreak (approved):

sebastianrakel (approved):

Issue #655, opened 2022-01-11
Add a share button in query


GermanG:
I'd like to share queries with the full URL, so I copy the query to the clipboard and share it like
http://puppetboard/query#inventory%5Bcertname%5D%7B%7D
And have the form populated ready to run the query (if not run automatically)

My javascript abilities are pretty limited, didn't find a way to do it what actually works, hence this issue.

Regards.

GermanG:
Made a simple change on the query.html to make it appear in the url, but I'm still not happy with the UX
```patch
--- a/puppetboard/templates/query.html
+++ b/puppetboard/templates/query.html
@@ -10,8 +10,29 @@
 {% endfor %}
 {% endif %}
 {% endwith %}
+
+<script type="text/javascript">
+  $(document).ready(function() {
+    if (window.location.href.indexOf("#") > -1) {
+        const pb_query = document.getElementById("query");
+        const hrefArray = window.location.href.split("#");
+        pb_query.value = decodeURI(hrefArray[1]);
+    }
+  });
+</script>
+
+<script language="JavaScript" type="text/JavaScript">
+    function chgAction()
+        {
+        const pb_form = document.getElementById("form");
+        const pb_query = document.getElementById("query");
+        pb_form.action = pb_form.action + "#" + encodeURI(pb_query.value);
+        pb_form.submit();
+        }
+</script>
+
 <div class="ui form">
-  <form method="POST" action="{{ url_for('query', env=current_env) }}">
+  <form method="POST" id="form" action="{{ url_for('query', env=current_env) }}">
     {{ form.csrf_token }}
     <div class="field {% if form.query.errors %} error {% endif %}">
       {{ form.query(autofocus="autofocus", rows=5, placeholder="nodes { certname = \"hostname\" }") }}
@@ -23,7 +44,9 @@
       <div class="field">
         {{ form.rawjson(active=True) }} Raw (but pretty and colored) JSON
       </div>
-      <input type=submit class="ui active submit button" value='Submit'>
+      <div  class="ui active submit button" value='Submit' onclick="chgAction()">
+          Submit
+      </div>
     </div>
   </form>
 </div>
```

gdubicki:
Released in [v3.4.0](https://github.com/voxpupuli/puppetboard/releases/tag/v3.4.0) today. Thanks again for your contribution here, @GermanG!

PR #657, opened 2022-01-18
Make the query shareable using URL

GermanG:
This is a PR with the code in #655

### DIFF
--- a/puppetboard/templates/query.html
+++ b/puppetboard/templates/query.html
@@ -10,8 +10,29 @@
 {% endfor %}
 {% endif %}
 {% endwith %}
+
+<script type="text/javascript">
+  $(document).ready(function() {
+    if (window.location.href.indexOf("#") > -1) {
+        const pb_query = document.getElementById("query");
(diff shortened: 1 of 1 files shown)


gdubicki:
Thanks for the PR!

Can you please share some screenshots for a quick glance how it looks, @GermanG ?

gdubicki:
LGTM! 👍 

I don't understand why you wrote [here](https://github.com/voxpupuli/puppetboard/issues/655#issuecomment-1014562475) that you don't like the UX though - what could we do better in your opinion?

GermanG:
I'm not completely happy with the UX because I'd like to have a "copy query" or "share" button to hit (so I won't need to select the URL myself) and the query is not submitted automagically when you visit the shared link.

gdubicki:
True, it could be improved. But the proposed changes won't break the compatibility with the existing solution and they can be done in a new PR so let's get this merged and released already. Thank you for your contribution, @GermanG!

PR #660, opened 2022-02-02
Configurable Binding host in Dockerfile

qhess34:
This small update aims to allow configuration of bind host on Docker.


gdubicki:
LGMT, thanks @qhess34! :)

Issue #669, opened 2022-03-01
Feature: Accept certificate and key as strings


SeanHood:
Hey folks, I'm setting up Puppetboard in AWS Fargate. I am running into a well known problem that in AWS Fargate you can't mount files to tasks (https://github.com/aws/containers-roadmap/issues/56).

So the feature I'm proposing is for the Puppetboard docker image to accept the certificate/key/ssl_verify files as strings via environment variables. Then taking those strings and writing them out to files.

Perhaps something like:
```
PUPPETDB_SSL_VERIFY_STRING
PUPPETDB_KEY_STRING
PUPPETDB_KEY_STRING
```

Or perhaps a variable to flag if they're strings or files.

Would there be interest in such a feature? I'm happy to attempt a PR if so.

gdubicki:
Hi @SeanHood!

Sure, we are open to PRs with such feature. Perhaps you could implement autodetection if the passed values are strings with certs or paths to such certs so we won’t have to introduce new env variables? I feel that we have too many already. 😅

PR #671, opened 2022-03-03
Support passing puppetdb certificates via environment variables

SeanHood:
Closes https://github.com/voxpupuli/puppetboard/issues/669

Adds support for passing certificates and keys to the following environment variables:
* PUPPETDB_SSL_VERIFY
* PUPPETDB_KEY
* PUPPETDB_CERT

I have tested this in my own environment as per the issue. With the container running in AWS Fargate, certificates being passed as env vars from AWS SSM Parameter store.

I had ran into a couple edge cases where the shell was breaking the new lines. These feel like general environment variable edge cases and not so much an issue with this code.

### DIFF
--- a/puppetboard/docker_settings.py
+++ b/puppetboard/docker_settings.py
@@ -1,4 +1,22 @@
 import os
+import tempfile
+
+
+def cert_to_file(cert_file_or_string):
+    """
+    cert_to_file takes in a string, if it looks like a certificate save the contents to a temporary
+    file and return the path to that temporary file
+
+    Note: NamedTemporaryFile does not work great with Windows, but since this is for Docker
+    hopefully we'll be fine.
+    """
+    if isinstance(cert_file_or_string, str) and '-----BEGIN' in cert_file_or_string:
+        with tempfile.NamedTemporaryFile(delete=False, suffix='puppetboard_cert') as tmpfile:
+            tmpfile.write(cert_file_or_string.encode())
+            tmpfile.close()
+        return tmpfile.name
+    else:
+        return cert_file_or_string
 
 
 def coerce_bool(v, default):
@@ -20,11 +38,11 @@ PUPPETDB_PORT = int(os.getenv('PUPPETDB_PORT', '8080'))
 # but if it is other string, then it's a path
 PUPPETDB_SSL_VERIFY = coerce_bool(
     os.getenv('PUPPETDB_SSL_VERIFY', True),
-    os.getenv('PUPPETDB_SSL_VERIFY')
+    cert_to_file(os.getenv('PUPPETDB_SSL_VERIFY'))
 )
 
-PUPPETDB_KEY = os.getenv('PUPPETDB_KEY', None)
-PUPPETDB_CERT = os.getenv('PUPPETDB_CERT', None)
+PUPPETDB_KEY = cert_to_file(os.getenv('PUPPETDB_KEY', None))
+PUPPETDB_CERT = cert_to_file(os.getenv('PUPPETDB_CERT', None))
 PUPPETDB_PROTO = os.getenv('PUPPETDB_PROTO', None)
 PUPPETDB_TIMEOUT = int(os.getenv('PUPPETDB_TIMEOUT', '20'))
 DEFAULT_ENVIRONMENT = os.getenv('DEFAULT_ENVIRONMENT', 'production')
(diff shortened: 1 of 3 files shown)


SeanHood:
That's a good shout, I'll add that to this PR.

As for where I've put this code does it make sense?

gdubicki:
> As for where I've put this code does it make sense?

Yes, for this purpose it makes perfect sense.

(Of course we could start a long design discussion that because it could be useful for non-dockerized Puppetboard, it should be moved out of `docker_settings.py`. Or that it should rather be implemented in the upstream [pypuppetdb](https://github.com/voxpupuli/pypuppetdb). Or that we should consider using something more secure than `/tmp` to keep the certificates (especially in non-dockerized env, with other processes possibly running on the host), so maybe we should use [memory-tempfile](https://pypi.org/project/memory-tempfile/) module and so on and on, but this would be overdesign at this point. We can get back to these ideas in the future, when more people will be interested in the feature. 😊)

gdubicki:
Please also add docs update, @SeanHood.

SeanHood:
> > As for where I've put this code does it make sense?
> 
> Yes, for this purpose it makes perfect sense.
> 
> (Of course we could start a long design discussion that because it could be useful for non-dockerized Puppetboard, it should be moved out of `docker_settings.py`. Or that it should rather be implemented in the upstream [pypuppetdb](https://github.com/voxpupuli/pypuppetdb). Or that we should consider using something more secure than `/tmp` to keep the certificates (especially in non-dockerized env, with other processes possibly running on the host), so maybe we should use [memory-tempfile](https://pypi.org/project/memory-tempfile/) module and so on and on, but this would be overdesign at this point. We can get back to these ideas in the future, when more people will be interested in the feature. 😊)

100% onboard with you there.

puppetboard/docker_settings.py
```diff
@@ -1,4 +1,34 @@
...
+            cert_str = base64.b64decode(cert_file_or_string).decode()
+        except (UnicodeDecodeError, binascii.Error):
+            cert_str = cert_file_or_string
+
+        if '-----BEGIN' in cert_str:
+            with tempfile.NamedTemporaryFile(delete=False, suffix='puppetboard_cert') as tmpfile:
+                tmpfile.write(cert_str.encode())
+                tmpfile.close()
+            return tmpfile.name
+        else:
```
russellcain, line 28: since you return on 27, you can drop this else and untab line 29, as that return of `cert_file_or_string` will just be hit if we aren't dealing with a `-----BEGIN` cert. 

side note, whenever I have an if/else with a large chunk of code between the two clauses, it can be a good time for an in-line comment saying `#  we know the user passed us a file name; lets return it directly`

puppetboard/docker_settings.py
```diff
@@ -1,4 +1,34 @@
...
+    """
+    if isinstance(cert_file_or_string, str):
+        try:
+            cert_str = base64.b64decode(cert_file_or_string).decode()
+        except (UnicodeDecodeError, binascii.Error):
+            cert_str = cert_file_or_string
+
+        if '-----BEGIN' in cert_str:
+            with tempfile.NamedTemporaryFile(delete=False, suffix='puppetboard_cert') as tmpfile:
+                tmpfile.write(cert_str.encode())
```
russellcain, line 25: do you want to be removing the `-----BEGIN CERTIFICATE-----` from the passed `cert_str` before you encode it? 
from the README, i kinda thought that the BEGIN and END would just be identifiers, but dropped by the time you used the actual cert.

russellcain (approved):

puppetboard/docker_settings.py
gdubicki: > since you return on 27, you can drop this else and untab line 29, as that return of `cert_file_or_string` will just be hit if we aren't dealing with a `-----BEGIN` cert.

True, but IMHO the current code is perfectly readable.

> side note, whenever I have an if/else with a large chunk of code between the two clauses, it can be a good time for an in-line comment saying `# we know the user passed us a file name; lets return it directly`

+1 to that!

puppetboard/docker_settings.py
gdubicki: We need to keep the BEGIN and END as they are part of the [PEM format](https://knowledge.digicert.com/quovadis/ssl-certificates/ssl-general-topics/what-is-pem-format.html) for cert/chain/key that the Requests library expects.

gdubicki:
Thanks for the code, @SeanHood, and for the code review, @russellcain!

PR #674, opened 2022-03-21
Add tests for friendly errors

gdubicki:


gdubicki:
Added to master as the main code is already in.

PR #675, opened 2022-03-24
Fix: handling facts in queries 

### LINKED ISSUE #658: Some regressions on v3.4.0 RC*
This is an early report, since I pulled from master.

In the Query section, API EndPoint PQL:
```
inventory [ certname, facts.osfamily ] { facts.osfamily = 'RedHat' }
```
This shows "Number of results" with a number, but only the headers are shown. 
Checking the JSON view the results are there.
If we only use 
```
inventory [ certname ] { facts.osfamily = 'RedHat' }
```
The table shows up, but certname is no longer a link to the node URL.

GermanG:
A dumb attempt to fix #658

### DIFF
--- a/puppetboard/templates/query.html
+++ b/puppetboard/templates/query.html
@@ -134,7 +134,7 @@
 'columns': [
     {% for column in columns %}
     {
-        "data": "{{ quote_columns_data(column) }}",
+        "title": "{{ quote_columns_data(column) }}",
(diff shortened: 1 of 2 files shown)


gdubicki:
If it works, it’s not dumb. :P

gdubicki:
Thanks a lot @GermanG, let’s try this out with a pre-release.

Issue #682, opened 2022-04-18
Query tab cannot be disabled on Docker


ldumont:
I'm running Puppetboard v3.4.1 using the official docker container running on Kubernetes using a home-made helm chart. 

I wanted to disable the Query tab using the "ENABLE_QUERY" environment variable but it does not work because the docker_settings.py file does not handle the variable as a boolean like other variables like OFFLINE_MODE or ENABLE_CATALOG (using the coerce_bool function).

The fix should be easy, I can try to provide a PR.

gdubicki:
Thanks for the fix @ldumont! It has been released in [v3.4.2](https://github.com/voxpupuli/puppetboard/releases/tag/v3.4.2) a moment ago.

PR #683, opened 2022-04-18
Fix  Query tab cannot be disabled on Docker #682

ldumont:
Use coerce_bool function for ENABLE_QUERY environment variable


gdubicki:
Thank you for the fix @ldumont!

Issue #687, opened 2022-05-23
Dockerfile healthcheck


jcpunk:
Some sort of built in health check that pulls down the "home page" would be very nice for testing DB connectivity.

gdubicki:
...but I merged #689 too fast, it's actually not working, @jcpunk.

Please see https://docs.python.org/3/library/sys.html#sys.exit - if you want to pass a non-integer to `sys.exit`, then it has to be `None` to exit with an exit code 0. And your expression returns a boolean. Can you please provide a working PR?

While fixing it please consider changing the `--interval` value too - 5 minutes means that together with the default `--retries=3` the app may not be working for 15 minutes until the health of the container will change to unhealthy - that's a pretty long time for such a low-cost healthcheck.

jcpunk:
Just posted a PR to try to clean up.

gdubicki:
Released in [v3.6.0](https://github.com/voxpupuli/puppetboard/releases/tag/v3.6.0). Thanks again for your contribution, @jcpunk!

PR #689, opened 2022-05-31
Minimal builtin health check

jcpunk:
This is related to #687 and adds a fairly minimal not that exciting health check.


gdubicki:
LGTM. Thanks for your contribution @jcpunk!

PR #690, opened 2022-06-01
Run health check more cleanly and more often

jcpunk:
In theory this should be right....


gdubicki:
Yeah, it works now.
```
CONTAINER ID   IMAGE         COMMAND                  CREATED              STATUS                        PORTS     NAMES
7a056c872cf2   puppetboard   "/bin/sh -c 'gunicor…"   About a minute ago   Up About a minute (healthy)   80/tcp    epic_raman
```
Thanks!

jcpunk:
Works for me :)

PR #692, opened 2022-06-06
Show puppetdb connection error message on homepage

ArthurWuTW:
- [ ]  Able to show error message on homepage when puppetdb connection failed.
image

- [ ]  Able to show 0 nodes on homepage without any agents yet.
image

### DIFF
--- a/puppetboard/app.py
+++ b/puppetboard/app.py
@@ -29,6 +29,7 @@ import puppetboard.views.radiator  # noqa: F401
 import puppetboard.views.reports  # noqa: F401
 # noinspection PyUnresolvedReferences
 import puppetboard.views.failures  # noqa: F401
+import puppetboard.errors
 
 
 from puppetboard.core import get_app, get_puppetdb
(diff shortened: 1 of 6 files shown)


gdubicki:
This implementation is definitely an improvement, but if you would use the [error handlers feature](https://flask.palletsprojects.com/en/1.1.x/errorhandling/#error-handlers) from Flask then it would work on any view, not just the Overview. Have you thought of that, @ArthurWuTW?

ArthurWuTW:
At first I wanted to make the minimal change to achieve those improvements for Overview only, but the error handler feature looks nice, I will check it out and give it a try :)

gdubicki:
> I just found that it has already had error handlers in code base.

Oups, I have really forgotten about it...

Well, anyway, it's great that we can achieve what we want with the minimal amount of changes that we have in this PR.

Can you also just please test if the Query view still behaves like it did? It's the only screen where we expect to return 400 or 404 in a normal day-to-day use, when a user enters an invalid query or it returns zero results. Please see https://github.com/voxpupuli/puppetboard/pull/653 for a reference, if needed.

If we still show the user-friendly messages there then I think we can merge and release this.

ArthurWuTW:
before this commit [71afad4](https://github.com/voxpupuli/puppetboard/pull/692/commits/71afad49ff371e8da3ce462232a22bcf7152f098), Query was like...
image


if we need to be able to query when 0 nodes, we need to add the same condition to skip env_check as Overview, and query behaves as normal.

image
image

gdubicki:
Thanks @ArthurWuTW, I think we are good to go with this now. :)

PR #693, opened 2022-06-07
failures breaks can break when no logs

carlosduelo:
A failed node that is not reporting any more can break failures raising the exception `UnboundLocalError: local variable 'source' referenced before assignment`

This should prevent this situation and show simple message


gdubicki:
LGTM, thanks @carlosduelo!

PR #697, opened 2022-06-25
V4 with Python 3.6, PuppetDBs < 5.2 support drop and Flask (and more) upgrade

gdubicki:
With a recent [first release of Requests that drops support for Python 3.6](https://github.com/psf/requests/releases/tag/v2.28.0) I realized that we should also drop it in the Puppetboard or else we will not be able to apply possible future security updates. Python 3.6 had its EOL in December 2021 so we already supported it for about 1/2 year extra.

The only well-known to me and officially maintained system that doesn't have Python 3.7+ is Centos 7, but for that I am hoping to provide a solution using Python 3.8 and HTTPD 2.4 from the open-suorce version of [Red Hat Software Collections 3.8 ](https://access.redhat.com/documentation/en-us/red_hat_software_collections/3/html/3.8_release_notes/chap-rhscl) downloaded from https://www.softwarecollections.org/en/. Or alternatively you can just use Docker.

As for the old PuppetDB versions drop: this is being done for code simplification and because we cannot reasonably claim that we support versions older than 5.2 if we don't have automatic acceptance tests that verify that the app works with them and we (or at least I) don't have the capacity to manually test old versions.

Finally, this is a nice opportunity to upgrade Flask and other core dependencies. We will have a few pre-releases of this major app version so hopefully we will root out the bugs created during this upgrade in these beta tests.


gdubicki:
We should probably combine it with #664...

README.md
```diff
@@ -26,8 +26,8 @@ See [more screenshots here](#more-screenshots).
 
 ## Requirements<a id="requirements"></a>
 
-* PuppetDB v. 3.0-7.5 (will most probably work with newer, but this has not been tested yet)
-* Python 3.6-3.10 or Docker
+* PuppetDB v. 5.2-7.5 (will most probably work with newer, but this has not been tested yet)
```
kenyon, line 29:
```suggestion
* PuppetDB v. 5.2-7.10 (will most probably work with newer, but this has not been tested yet)
```

puppetboard works for me with puppetdb 7.10.

gdubicki:
Hm, not sure why the tests are passing for me locally but not in GHA... Do I have to mock the response from the PuppetDB version endpoint for all the tests? 🤔

gdubicki:
Please let me know what you think about this, @kenyon @bastelfreak @sebastianrakel @mterzo. Is the scope ok? Or should we add some other backward-incompatible change while we are doing major uptick? (Ping more people if needed, of course.)

puppetboard/utils.py
```diff
@@ -31,27 +32,40 @@ def jsonprint(value):
...
-        log.info("PuppetDB Version %d.%d.%d" % (major, minor, build))
-    except ValueError:
-        log.error("Unable to determine version from string: '%s'" % puppetdb.current_version())
-        ver = (4, 2, 0)
+
+        log.info(f"PuppetDB version: {version}")
+
+        # Puppet Server is enforcing new metrics API (v2)
+        # starting with versions 6.9.1, 5.3.12, and 5.2.13
+        if parse('5.2.0') <= parse(version) < parse('5.2.13'):
```
ekohl, line 46: Nit: This is calling `parse(version)` quite a few times. Perhaps it should be stored in a variable?

gdubicki: Thanks @ekohl 🙏, corrected. 😊

gdubicki:
Please check out a pre-release [v4.0.0b1](https://github.com/voxpupuli/puppetboard/releases/tag/v4.0.0b1).

(Note: because of a PyPI issue I had to install it from the URL https://files.pythonhosted.org/packages/09/9d/5e59c13b3c9bb4ab250440b7fb1fd7385ece9213946983de41ac2e186ff0/puppetboard-4.0.0b1-py2.py3-none-any.whl )

gdubicki:
Ping also @smortex and @buzzdeee as the maintainers of Puppetboard packages.

bastelfreak (approved):

gdubicki:
I see no objections and I have an approval (thanks @bastelfreak!), so I am moving forward with this.

PR #699, opened 2022-06-29
Remove border-radius from Top Menu

promarcel:
From a user experience perspective, it would make sense to remove the `border-radius` which is added by `semantic.min.css` and thus make the top navigation appear more clearly anchored to the page.

### DIFF
--- a/puppetboard/static/css/puppetboard.css
+++ b/puppetboard/static/css/puppetboard.css
@@ -112,6 +112,7 @@ h1.ui.header.no-margin-bottom {
 
 .ui.menu.darkblue {
   background-color: #2C3E50;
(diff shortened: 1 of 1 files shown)


gdubicki:
Thanks for the contribution @promarcel! It does look better without those rounded corners. 😊

gdubicki:
Oops, I forgot to mention you in the release that included your contribution - I fixed it now, please check out the [v4.0.1](https://github.com/voxpupuli/puppetboard/releases/tag/v4.0.1) [changelog entry](https://github.com/voxpupuli/puppetboard/blob/master/CHANGELOG.md#401).

PR #705, opened 2022-07-15
Improve UI of the environment selection

### LINKED ISSUE #368: Limit environment drop down to a subset
Add the ability to limit he environment drop down.

Could be a static list to start with maybe, add the ability to regex or pattern matching.

melck:
Hi,

I fixed an old issue which concerns the selection of environments. When you have many environments, display and selection are quite difficult.

I changed the location to the right of the menu and used the Semantic UI select search dropdown to make it easier to search and select.

However I had to add some javascript to do the redirection and css to fix dropdown / loading icon.

The only inconvenience of this solution is that you can no longer right click > open in new tab on an environment.

I tested this fix in the latest chrome and firefox browsers.

Example of the issue :

puppetboard-environment-issue

Results after fix :

puppetboard-environment-result

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,8 +155,11 @@ For information about how to generate the correct keys please refer to the
 [pypuppetdb documentation](https://pypuppetdb.readthedocs.io/en/latest/connecting.html#ssl). Alternatively it is possible
 to explicitly specify the protocol to be used setting the `PUPPETDB_PROTO` variable.
 
-Other settings that might be interesting in no particular order:
+Other settings that might be interesting, in no particular order:
 
+- `FAVORITE_ENVS`: an ordered list of Puppet environment names that will be shown immediately after "All Environments"
+    and before other environments (which are sorted by name) in the dropdown for choosing the environment shown
+    in the top-right of the UI. Environments listed here that do not really exist in your deployment are silently ignored.
 - `SHOW_ERROR_AS`: `friendly` or `raw`. The former makes Puppet run errors in Report and Failures views shown
     in a modified, (arguably) more user-friendly form. The latter shows them as they are.
     Defaults to `friendly`.
--- a/puppetboard/core.py
+++ b/puppetboard/core.py
@@ -75,15 +75,53 @@ def get_puppetdb():
     return PUPPETDB
 
 
-def environments():
+def environments() -> dict:
+    envs = {}
     puppetdb = get_puppetdb()
-    envs = get_or_abort(puppetdb.environments)
-    x = []
+    envs_from_db = sorted(
+        env['name']
+        for env in get_or_abort(puppetdb.environments)
+    )
 
-    for env in envs:
-        x.append(env['name'])
+    # Adding all environments
+    envs['All Environments'] = {
+        'url': url_for_field('env', '*'),
+        'icon': 'server',
+        'divider': True
+    }
 
-    return x
+    # Adding favorite envs
+    favorite_envs = [
+        env
+        for env in get_app().config.get('FAVORITE_ENVS')
+        if env in envs_from_db
+    ]
+
+    divider = True
+    for env in favorite_envs:
+        envs[env] = {
+            'url': url_for_field('env', env),
+            'icon': 'star',
+            'divider': divider
+        }
+        # Only the first envs from this group should have a divider
+        divider = False
+
+    # Adding other envs
+    divider = True
+    for env in envs_from_db:
+        if env in favorite_envs:
+            continue
+
+        envs[env] = {
+            'url': url_for_field('env', env),
+            'icon': '',
+            'divider': divider
+        }
+        # Only the first envs from this group should have a divider
+        divider = False
+
+    return envs
 
 
 # as documented in
--- a/puppetboard/default_settings.py
+++ b/puppetboard/default_settings.py
@@ -59,5 +59,6 @@ FAVORITE_ENVS = [
     'production',
     'staging',
     'qa',
+    'test',
     'dev',
 ]
--- a/puppetboard/default_settings.py
+++ b/puppetboard/default_settings.py
@@ -53,6 +53,11 @@ REFRESH_RATE = 30
 DAILY_REPORTS_CHART_ENABLED = True
 DAILY_REPORTS_CHART_DAYS = 8
 WITH_EVENT_NUMBERS = True
-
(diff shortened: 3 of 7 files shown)


gdubicki:
Hi @melck!

Thank you for your contribution! It does make choosing an env easier when you have a long list of them and it looks really nice.

However I don't think that it fixes #368 as that ticket is - in my opinion - about something a bit different. I think that #368 is a request to add a configuration setting that will allow to _permanently_ limit the elements shown in the environment chooser, f.e. to show only your main environments like: ["production", "staging", "dev"].

I am not trying to downplay your work, of course, I just want to put the right info in the changelog when we merge and release this. :)

What do you think?

gdubicki:
(The force push above was to rebase to include a fix from v4.0.2)

Can we make the new dropdown also expand up to the height of the screen, @melck? Because if you have a long list of envs and want to use the old way of just clicking on the right env, without using the search, it now requires you to do an extra scroll.

melck:
Reading it again, I see what you mean  for ticket #368. I will rebase to change my commit message. What do you prefer @gdubicki as commit message for the changelog ?

Conerning the old ways for the dropdown and in my case, we have so many envs that we need to scroll down to get env and sometime the scroll down make disapear the dropdown. So we modify url directly to see specific env. 

I will try to surchage the CSS for this dropdown and do a mix of two world by changing max height of visible dropdown menu.

gdubicki:
Thanks for a quick response!

In the meantime I did some hacking to try to implement what I consider the gist of #368: to make selected envs shown as the first on the list. Please check out my draft PR to your branch here: https://github.com/melck/puppetboard/pull/1 and let me know what you think. :) (It's a draft as it would require completing the migration from Semantic to Fomantic in the offline mode and some more testing to make sure that all views work and look fine after the migration.)

melck:
@gdubicki I have made some changes from you proposals. I have refactor some code to have a better front visibility beetween using Formantic dropdown divider.

I force order of environments by first using current env, then envs from settings, then others envs. This allow to always see first envs and avoid to scroll for a simple usage.

Like you made me update to Fomantic. I updated all js dependencies and i dropped c3js for billboard (who seems maintained). I fix dailycharts to use billboard and remove some code to avoid multiple call of the script.

What do you think ?

gdubicki:
Wow @melck, that's a fantastic contribution! Frankly I didn't expect you to update all those dependencies, I assumed that I would have to do it... It's great that I didn't have to. 😅

Let me run some tests and review the code.

bastelfreak:
@melck can you please provide https://github.com/voxpupuli/puppetboard/pull/705/commits/f108e1c764bd5ba896a65f467dd4146fe2b33e75 as a dedicated PR and maybe document the update process (is there maybe a bot that we could use for that)

melck:
> @melck can you please provide [f108e1c](https://github.com/voxpupuli/puppetboard/commit/f108e1c764bd5ba896a65f467dd4146fe2b33e75) as a dedicated PR and maybe document the update process (is there maybe a bot that we could use for that)

Ok for that i was thinking of making a separate jinja template for css / js and creating a setup subcommand, what do you think ?

gdubicki:
@melck: I think that the new env dropdown would look even nicer if it would not introduce yet another color to the UI.

Do you know how to change the background color of the selected dropdown to the same as active main menu tab? I have tried that and failed, my frontend skills are not that great...

Screenshot 2022-07-17 at 10 56 42

melck:
> @melck: I think that the new env dropdown would look even nicer if it would not introduce yet another color to the UI.
> 
> Do you know how to change the background color of the selected dropdown to the same as active main menu tab? I have tried that and failed, my frontend skills are not that great...

Yeah last Fomantic UI added a color when menu gained state :hover. I have fixed this issue for inverted menu.

melck:
@gdubicki I have another suggestion for the python part, do you know Black?

It is a Python code formatter that allows you to always have the same rendering. It's easier to read python code formatted by Black when you adopted it. I think it could be interresting at some time to apply it everywhere.

What do you think ?

gdubicki:
> I force order of environments by first using current env, then envs from settings, then others envs. This allow to always see first envs and avoid to scroll for a simple usage.

Actually I am not a fan of changing the dropdown content based on what is selected. I think it results in bad UX as the user doesn't expect that...

How about we always show:
1. All Envs
2. (divider)
3. First envs (as configured)
4. (divider)
5. Other envs (sorted)

? 

Also I like the idea of adding icons, but let's make them meaningful. For example use a different one for all envs, first envs and the rest.

It could look like this: 

Screenshot 2022-07-17 at 11 52 28

What do you think?

gdubicki:
> @gdubicki I have another suggestion for the python part, do you know Black?
> 
> It is a Python code formatter that allows you to always have the same rendering. It's easier to read python code formatted by Black when you adopted it. I think it could be interresting at some time to apply it everywhere.
> 
> What do you think ?

I am totally for but definitely in a separate PR. 😅

melck:
> > I force order of environments by first using current env, then envs from settings, then others envs. This allow to always see first envs and avoid to scroll for a simple usage.
> 
> Actually I am not a fan of changing the dropdown content based on what is selected. I think it results in bad UX as the user doesn't expect that...
> 
> How about we always show:
> 
> 1. All Envs
> 2. (divider)
> 3. First envs (as configured)
> 4. (divider)
> 5. Other envs (sorted)
> 
> ?
> 
> Also I like the idea of adding icons, but let's make them meaningful. For example use a different one for all envs, first envs and the rest.
> 
> What do you think?

Yes it was an hacking way to avoid autoscroll to environment selected and not seeing the fixed one, but i have another idea to try. I think dropdown can have fixed headers and the scroll will be only on "others envs", what do you think ?

gdubicki:
> Yes it was an hacking way to avoid autoscroll to environment selected and not seeing the fixed one, but i have another idea to try. I think dropdown can have fixed headers and the scroll will be only on "others envs", what do you think ?

Hack away. :) I have pushed my proposal but feel free to revert or change it as you want to show your solution.

melck:
@gdubicki I have rebased code with master since you merge my other PR.

bastelfreak (approved):

gdubicki:
Thank you for all your work on this @melck and for your review, @bastelfreak! After adding basic docs I think it's ready to merge. :)

Issue #706, opened 2022-07-22
Facts with int values which are strings internally are not findable


ryaner:
**Describe the bug**

Currently puppetboard is able to detect the facts value type correctly from the puppetdb calls, but the generated links do not differentiate between string and int, resulting in the call to `parse_python` switching it back to an int.

An example will hopefully make this clearer

Using Centos operating system facts, which are stored into puppetdb as strings.

https://puppetdb/fact/operatingsystemmajrelease
This page will list all the hosts with this fact present. The generated html is showing type string here too
```
<td><a title="6" href="/fact/operatingsystemmajrelease/6"><span class="string">6</span></a></td>
```
Following the link - https://puppetdb/fact/operatingsystemmajrelease/6 - will display no entries.

The facts code is calling `parse_python` which is returning the `int` type to the query.
https://github.com/voxpupuli/puppetboard/blob/c0db5abcc09e419381b457af45bda2dd543cc621/puppetboard/views/facts.py#L62-L65

If I manually quote the value - https://puppetdb/fact/operatingsystemmajrelease/"6" - the search works and displays results.

Values which are `int` internally, and displayed as `int` on the facts page *do* search correctly.

**Puppetboard version**

F.e. 3.6.1 (verified code has not changed for this flow since)

**Environment and installation method**

F.e.:
* Centos 7 and Python 3.8,
* PyPI package.
* puppetdb-6.21.0-1

While quoting all strings does work as a solution here, that just feels wrong which is why I haven't attached a PR here. If you have a better solution that could work, I'm happy to PR things.
Pretty print explicitly tries to not quote things too https://github.com/voxpupuli/puppetboard/blob/c0db5abcc09e419381b457af45bda2dd543cc621/puppetboard/static/js/pretty.js#L5-L6

gdubicki:
Hi @ryaner! Thanks for reporting this. 

I don’t know when I will have the time to dig into this as I will be pretty busy over the next ~2 weeks so even more than always, PRs with a fix is very welcome.

ryaner:
My current patch is below, but I keep thinking there must be a better solution that this. So far it does work on all the `int` stored as `string` values though.
```
diff --git a/puppetboard/views/facts.py b/puppetboard/views/facts.py
index 714eba9..2715cd0 100644
--- a/puppetboard/views/facts.py
+++ b/puppetboard/views/facts.py
@@ -93,7 +93,7 @@ def fact_ajax(env, node, fact, value):
[...]

PR #707, opened 2022-07-25
Force quote strings for the facts page - #706

ryaner:


gdubicki:
Thanks for the PR, @ryaner!

PR #708, opened 2022-07-25
Update frontend dependencies

melck:
Upgrade frontend dependencies :
- Move external frontend dependencies to `puppetboard/static/libs`
- Move external frontend dependencies from `layout.html` to `_offline_static.html` and `_online_static.html`
- Upgrade fontrend dependencies
- Replacing jquery tablesort by datatable.net for all sortable tables
- Replacing c3.js by billboard.js and fix charts
- Fix loading buttons for datatables.net 
- Enabling jinja stip blocks to remove block from final HTML

Move to another PR after discussions
- Layout : 
  - Clean multiple block for loading javascript
  - Change content page on 15 column centered
  - Change menu to be fixed
- Trying to visually improve index page

@gdubicki This draft PR is from discussion on #705 with some visual changes. What do you think of it ?

PR #710, opened 2022-09-09
Fix frontend dependencies

melck:
Remove unwanted changes from #708 :
- Regression for nodes order on index page 

@gdubicki I have missed some localized timestamp fields. I have add them to this PR


gdubicki:
Great, thank you @melck!

PR #712, opened 2022-09-12
feature: improve os detection for inventory page

### LINKED ISSUE #485: Inventory OS column blank on all nodes running Centos 7
I have 7 nodes running CentOS and 1 running Ubuntu and none of the CentOS nodes are displaying a value in the OS column on the inventory screen.
Puppet: 6.0
PuppetDB: 6.0
puppetboard: 0.3.0, installed manually and running on Apache with Mod-WSGI

melck:
Hello,

I wanted to use inventory page but it only works with few linux OS. Here is my proposal to support many OS. I have introduced a new parameter named `INVENTORY_FACT_TEMPLATES` to keep customization. This parameter is used to customize cell display using jinja. Default INVENTORY_FACTS has been changed to use trusted and os facts.

 It has been tested with Puppet agent >= 5 to 7:
- Windows (Pro, Server 2012/2016/2019)
- MacOS
- Debian Family (Debian, Ubunut, Linux Mint, etc.)
- Redhat Family (RedHat Enterprise, Oracle Linux, CentOS)
- Suze Family (SUZE Enterprise)

With puppet agent 4, windows os will display nothing.

It should fix #485. 

What do you think about this changes ?

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,7 +122,7 @@ The file has to be identical to
 [default_settings.py](https://github.com/voxpupuli/puppetboard/blob/master/puppetboard/default_settings.py)
 but should only override the settings you need changed.
 
-If you run PuppetDB and Puppetboard on the same machine the default settings provided will be enough to get you started 
+If you run PuppetDB and Puppetboard on the same machine the default settings provided will be enough to get you started
 and you won't need a custom settings file.
 
 Assuming your webserver and PuppetDB machine are not identical you will at least have to change the following settings:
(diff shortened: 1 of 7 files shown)


bastelfreak (commented):
to me the change looks fine, but someone with more Python knowledge should take a look as well. @sebastianrakel could you take a look?

README.md
```diff
@@ -186,6 +186,7 @@ Other settings that might be interesting, in no particular order:
     values being unique per node, like ipaddress, uuid, and serial number, as well as structured facts it was no longer
     feasible to generate a graph for everything.
 - `INVENTORY_FACTS`: A list of tuples that serve as the column header and the fact name to search for to create
+- `INVENTORY_FACT_TEMPLATES`: A mapping between fact name and jinja template to customize display
```
gdubicki, line 189: This line should probably be added one line lower.

gdubicki (changes requested):
I found some time. :) The MR looks good to me except one minor comment about the docs.

gdubicki:
I have pre-released this in [v4.1.0rc3](https://github.com/voxpupuli/puppetboard/releases/tag/v4.1.0rc3). It would be great if we could add a screenshot of the Inventory view showing nodes with various OSes for the final v4.1.0 release that I would like to do in a few days. Can you provide such screenshot, @melck? I don't have nodes with other OS than Linux in my infrastructure...

PR #715, opened 2022-09-24
Restore facts filtering feature

### LINKED ISSUE #714: Filtering doesn't work on Facts view
**Describe the bug**

When you enter some text in the "type here to filter" text field in Facts view the facts below are not filtered - nothing happens.

**Puppetboard version**

4.1.0rc2

(Note: this bug does not occur in 4.0.5.)

**Environment and installation method**

Centos 7 and Python 3.8 from Software Collections, installed with the Forge module and using the git repo tag.

gdubicki:
that has been removed in https://github.com/voxpupuli/puppetboard/pull/708

Fixes #714


gdubicki:
See https://github.com/voxpupuli/puppetboard/pull/716 instead.

gdubicki:
Effectively these are only @melck’s changes which I approve, so I think we are good to go with this PR. :)

PR #716, opened 2022-09-26
Restore and refactor facts filtering

### LINKED ISSUE #714: Filtering doesn't work on Facts view

melck:
that has been removed in #708

Fixes #714


gdubicki:
Cool, it’s good to get rid of a technology if it’s used for one thing only. :) 

So it will be functionally the same as before?

gdubicki (changes requested):
Can you please also remove coffeescript from dev.py and possibly other references?

melck:
Hello @gdubicki ,

I got rid of all related coffeescript things. Since the file `dev.py` was only useful to handle coffeescript compilation, i removed it to use default flask command. I updated dev docs in consequences.

Regarding functionnality, it will be the same but i have added a sightly delay (250ms) for filtering to avoid weird freeze of browsers when typing to fast.

gdubicki (approved):

Issue #461, opened 2018-04-13
JS broken when LOCALISE_TIMESTAMP set to False


volans-:
In [_macros.html#L71](https://github.com/voxpupuli/puppetboard/blob/master/puppetboard/templates/_macros.html#L71) there is a call to `localise_timestamp()` that is done independently of the `LOCALISE_TIMESTAMP` setting. When `LOCALISE_TIMESTAMP` is set to `False`, `moment.js` is not loaded at all (see `layout.html`) resulting in JS errors:
```
Uncaught TypeError: $(...).localise_timestamp is not a function
```
Ideally even when `LOCALISE_TIMESTAMP` is set to `False` the timestamps should be properly formatted in a nice way, just preserving the original times without TZ conversion.

As a side note the `README` is a bit confusing regarding this option given that it states:

> LOCALISE_TIMESTAMP: Normalize time based on localserver time.

While it seems to me that it normalize it to the client's time when set to `True`.

emlun007:
moment.js should be loaded in any case to get proper formatting of the data:

```
{% if config.OFFLINE_MODE %}

      <script src="{{ url_for('static', filename='jquery-2.1.1/jquery.min.js') }}"></script>
      <script src="{{ url_for('static', filename='jquery-datatables-1.10.13/jquery.dataTables.min.js') }}">
      <script src="{{ url_for('static', filename='jquery-datatables-1.10.13/dataTables.semanticui.min.js') }}">
      <script src="{{ url_for('static', filename='moment.js-2.7.0/moment.min.js') }}"></script>

```

So to get a nice formatting, just without TZ conversion, in timestamps.js you can just remove the line:
`      result.local();`

to get the data in utc, so something like: 

```
$.fn.extend({
  localise_timestamp: function (){
    var tstring = $(this).text().trim();
    if (tstring === "None"){
      $(this).text('Unknown');
    } else {
      var result = moment(tstring).utc();
      $(this).text(result.format('MMM DD YYYY - HH:mm:ss'));
    }
  }
})
```

gdubicki:
Hi @volans-!

The code you mentioned has been changed in https://github.com/voxpupuli/puppetboard/pull/708 but I noticed today that `LOCALISE_TIMESTAMP=False` still did not work.

But I created https://github.com/voxpupuli/puppetboard/pull/722 that should make it work.

volans-:
@gdubicki Thanks for the follow up.

gdubicki:
[...]

PR #722, opened 2022-10-23
Fix LOCALISE_TIMESTAMP=False and document better how LOCALISE_TIMESTAMP works

gdubicki:
Fixes https://github.com/voxpupuli/puppetboard/issues/461

Plus various fixes for running tests.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,7 +32,6 @@ coverage.xml
 .tox
 .mypy_cache
 .pytest_cache
-nosetests.xml
 
 # Translations
 *.mo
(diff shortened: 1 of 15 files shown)


.gitignore
```diff
@@ -32,7 +32,6 @@ coverage.xml
 .tox
 .mypy_cache
 .pytest_cache
-nosetests.xml
```
bastelfreak, line 35: why did you remove it from .gitignore? The file isn't part of this commit.

gdubicki: Because we are not using nosetests in this project.

gdubicki: I noticed it while I was fixing flake8 issues.

gdubicki: Is that ok, @bastelfreak?

bastelfreak (approved):

.gitignore
gdubicki: Thank for the approval, @bastelfreak !

PR #985, opened 2023-12-01
Update README for SECRET_KEY and SSL Setup

tuxmea:
- add section on SECRET_KEY
- add example for PuppetDB SSL within Puppet code
- correct parameter explanation

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ To see how to get it working with RedHat/Centos 7 check out these [docs](https:/
 
 We provide [an official Docker image in the GitHub Container Registry](https://github.com/orgs/voxpupuli/packages/container/package/puppetboard).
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -46,12 +46,15 @@ To see how to get it working with RedHat/Centos 7 check out these [docs](https:/
 
 We provide [an official Docker image in the GitHub Container Registry](https://github.com/orgs/voxpupuli/packages/container/package/puppetboard).
 
+You must provide a secret key. Generate one by running `ruby -e "require 'securerandom'; puts SecureRandom.hex(32)"`.
```
gdubicki, line 49: Nit: I would suggest wording like "Generate one for example by running" to prevent assumptions that this command is the only way to generate a secret key.

gdubicki (approved):
One minor comment, but LGTM in general! Thank you for the contribution, @tuxmea!

bastelfreak (approved):

PR #1023, opened 2024-02-07
Add favicon to radiator

surdaft:
I see 404 errors in nginx due to the radiator looking for `/favicon.ico` instead of my subfolder url that the rest of puppetboard looks for.

Small fix, but reduces a little bit of noise for me. 😅


bastelfreak:
thanks for the fix!

PR #1027, opened 2024-02-08
test trivy execution

rwaffen:


rwaffen:
hmm strange... trivy is executed, but the results show not up. and it is exactly configured like in puppetserver or hdm :thinking:

PR #1086, opened 2024-04-14
Align structured facts with tops of cells

yakatz:
Structured facts are potentially very long, so it can be hard to find the keys and values. This puts them at the top of the cells, which might be less visibly pleasing from a design perspective, but is definitely more useful.

### DIFF
--- a/puppetboard/static/css/puppetboard.css
+++ b/puppetboard/static/css/puppetboard.css
@@ -275,3 +275,7 @@ td.code {
 .dataTables_length > label {
   margin-right: 1em;
 }
+
+#facts_table td {
+  vertical-align: top;
+}


yakatz:
On a `node` page:

Before
image

After
image

yakatz:
On a `fact` page:

Before
image


After
image

smortex (approved):
Slightly better.  Top-alignment of the fact names make sense as some facts are very long and this makes us be sure we can scroll up to read the fact name.

Maybe we improve this further by limiting the height used to display a fact value, maybe "wrapping" long facts values and allowing the user to unwrap them form the UI with a button (look complicated), or limiting the value max height and having a scroll-bar on overflow (full CSS); but it is maybe something for another PR.

yakatz:
> Maybe we improve this further by limiting the height used to display a fact value, maybe "wrapping" long facts values and allowing the user to unwrap them form the UI with a button (look complicated), or limiting the value max height and having a scroll-bar on overflow (full CSS); but it is maybe something for another PR.

If I ever have time, I was thinking about looking into some javascript that could make the structured facts collapsible - no idea when I would get to it...

gdubicki:
Thank you for the contribution, @yakatz!

Feel free to create your PR to further improve this, @smortex and @bastelfreak 😊

Issue #988, opened 2023-12-05
Offline mode tries to load external resources


teluq-pbrideau:
**Describe the bug**

Bug #459 from 2018 appear to have come back:

When in offline mode (OFFLINE_MODE set to True), Puppetboard should not load external resources.

First line in https://github.com/voxpupuli/puppetboard/blob/master/puppetboard/static/libs/fomantic-ui/semantic.min.css
`@import url(https://fonts.googleapis.com/css2?family=Lato:ital,wght@0,400%3B0,700%3B1,400%3B1,700&subset=latin&display=swap);`

**Puppetboard version**

5.1.0

**Environment and installation method**

* Rocky8, Python 3.9
* Installed with Forge module v9.0.0 and using PyPI package from an internal mirror

teluq-pbrideau:
The problems seems to have solved itself…  Sorry for the false alarm.

I guess there was big timeout somewhere in the network trying to resolve `fonts.googleapis.com.` before proceeding to display the page?  Now, there is no more timeout, everything work as expected.  Maybe the network team was doing a DROP for this request and now it is a REJECT which fail fast?  Just a guess, but my problem is now solved.

teluq-pbrideau:
Sorry for the opening and closing this issue, the Network team had open the access of the client to the internet by error, so obviously it was working…  Now they have closed and the problem is back again…

I’ve tried to debug and change something like #459 had suggested:

```
url(https://fonts.googleapis.com/css2?family=Lato:ital,wght@0,400%3B0,700%3B1,400%3B1,700&subset=latin&display=swap);
```
to
```
@import url(/static/css/fonts.css);
```
I got an error, the web page layout is completely out of wack, but I does not take 1 minute to load the page anymre…  It seems to be really where the problem is…  But I’m no web developper, so I don’t really know how to fix this problem…

This is the error in the web console.  My change seems to have broken a checksum somewhere…
```
None of the “sha512” hashes in the integrity attribute match the content of the subresource. The computed hash is “C1tpQ9+Wmn2KscdEXRUVo7hnmiHRgR9xZqMWwV4VZ1mnyTAC7AV2stGG7VgNs5dz+r8pwKblfAeOHrPpnPmjOQ==”.
```

kenyon:
https://github.com/google/fonts#self-host-fonts-available-from-google-fonts -> https://fontsource.org/docs/getting-started/install

[...]

PR #1124, opened 2024-06-27
Fix offline

VolJin:
Fix for https://github.com/voxpupuli/puppetboard/issues/988
When using offline mode, the puppetboard tries to load external resources.


yakatz:
Manually changing the included CSS from another library doesn't seem like a good idea - it will get overwritten next time the library is updated. Adding an NPM build is probably the better option.

PR #1202, opened 2024-11-26
CI: Test if we can build puppetboard & install setuptools

bastelfreak:
In our normal test pipeline, we didn't verify if we can properly build the package. because of that, the last 6.0.0 release failed:

```
Run python setup.py build sdist bdist_wheel
Traceback (most recent call last):
  File "/home/runner/work/puppetboard/puppetboard/setup.py", line 3, in <module>
    from setuptools.command.test import test as TestCommand
ModuleNotFoundError: No module named 'setuptools'
```

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -13,7 +13,12 @@ jobs:
 
     strategy:
       matrix:
-        python-version: [3.9, '3.10', 3.11, 3.12, 3.13]
+        python-version:
+        - 3.9
+        - '3.10'
+        - 3.11
+        - 3.12
+        - 3.13
 
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


ekohl (commented):
I'm wondering if we need to declare setuptools as a dependency in some way. Or migrate to `pyproject.toml`. https://packaging.python.org/en/latest/guides/writing-pyproject-toml/ says we can declare setuptools as the build system:
```toml
[build-system]
requires = ["setuptools >= 61.0"]
build-backend = "setuptools.build_meta"
```

No idea if we really need `>= 61.0` or any version.

bastelfreak:
> I'm wondering if we need to declare setuptools as a dependency in some way

I do not know enough about python.

d1nuc0m:
Hi @vchepkov, could you review this one instead of #1204 ? Thanks

vchepkov (approved):

PR #1206, opened 2024-11-29
Add APPLICATION_ROOT to docker_settings.py

### LINKED ISSUE #1092: The PUPPETBOARD_URL_PREFIX does not set flask's APPLICATION_ROOT
To whom it may concern:

By adding the following

APPLICATION_ROOT = os.getenv('PUPPETBOARD_URL_PREFIX','/')

to the [docker_settings.py](https://github.com/voxpupuli/puppetboard/blob/master/puppetboard/docker_settings.py)

and creating a new docker image the puppetboard page work for me. 

Very cool project!

Regards, Andrew

HielkeJ:
Add APPLICATION_ROOT to docker_settings.py. This way setting PUPPETBOARD_URL_PREFIX works. If APPLICATION_ROOT is not set and PUPPETBOARD_URL_PREFIX is set, every call will result in a 500 error with "IndexError: list index out of range".

This should resolve #1092

### DIFF
--- a/puppetboard/docker_settings.py
+++ b/puppetboard/docker_settings.py
@@ -45,6 +45,7 @@ def coerce_bool(v, default):
         return False
     return default
 
+APPLICATION_ROOT = os.getenv('PUPPETBOARD_URL_PREFIX','/')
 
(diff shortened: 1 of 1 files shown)


bastelfreak:
thanks for the fix!

PR #1242, opened 2025-02-13
Release 6.0.1

rwaffen:
- run pipeline again to publish
- switch version detection in changelog to branch name

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,9 @@
 
 All notable changes to this project will be documented in this file.
 
-## [6.0.1](https://github.com/voxpupuli/puppetboard/tree/6.0.1) (2025-02-13)
+## [v6.0.1](https://github.com/voxpupuli/puppetboard/tree/v6.0.1) (2025-02-13)
 
-[Full Changelog](https://github.com/voxpupuli/puppetboard/compare/v6.0.0...6.0.1)
+[Full Changelog](https://github.com/voxpupuli/puppetboard/compare/v6.0.0...v6.0.1)
 
 There were some pipeline issues for the 6.0.0, so it's not available on pypi. The 6.0.1 release is almost identical, it just updates some dependencies and switches to poetry (which fixed the CI).
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,25 @@
 
 All notable changes to this project will be documented in this file.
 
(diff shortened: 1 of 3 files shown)


Rakefile
```diff
@@ -4,8 +4,7 @@ rescue LoadError
   # github_changelog_generator is an optional group
 else
   GitHubChangelogGenerator::RakeTask.new :changelog do |config|
-    version = File.read('puppetboard/version.py').lines.last.scan(/(\d+\.\d+\.\d+)/).first.first
-    config.future_release = "v#{version}" if /^\d+\.\d+.\d+$/.match?(version)
+    config.future_release = `git rev-parse --abbrev-ref HEAD`.strip.split('-', 2).last
```
bastelfreak, line 7: This just parses the branch name? Not sure how reliable this is. for poetry, we could use `poetry version -s` (but then we need poetry installed :(

rwaffen: this is how it works in my other repos when you stick to the brnach name pattern from RELEASE.md ... and peotry updates only the toml file, afaik. and the rake task ist still broken then because in the version.py is not version aynmore :shrug:

bastelfreak:
```ruby
File.readlines('pyproject.toml').find{|l| l.match(/version/)}.split(' ').last.gsub('"', '')
```
also not very beautiful? :D

rwaffen: If you like it this way. Don’t care. Just want a release. But also like my solution more. But won’t fight over it.

bastelfreak:
I created a milestone with a release summary: https://github.com/voxpupuli/puppetboard/milestone/8?closed=1

can you please regenerate the changelog? It should then contain the release summary (which explains the missing 6.0.0 release).

bastelfreak (approved):

CHANGELOG.md
```diff
@@ -2,6 +2,27 @@
 
 All notable changes to this project will be documented in this file.
 
+## [6.0.1](https://github.com/voxpupuli/puppetboard/tree/6.0.1) (2025-02-13)
```
bastelfreak, line 5: I just noticed that previews releases were prefixed with v, we should keep that?

bastelfreak (changes requested):
I just noticed that previews releases were prefixed with v, we should keep that?

CHANGELOG.md
rwaffen: urgs, yes, please, will update, mom

bastelfreak (approved):

PR #1292, opened 2025-06-04
feat: Replace poetry with uv

nikolaik:
I'm tentatively opening this PR to see if there is interest in moving to uv from poetry

- Update README.md
- Update GHA workflows
- Regenerate requirements.txt


yakatz:
Thanks for the PR. What would be the motivation for the change?

nikolaik:
> Thanks for the PR. What would be the motivation for the change?

Hi, the motivation for the change would be to improve the ergonomics of working with the repo. Main point being faster installs and venv management. The list of highlights on the [uv website](https://docs.astral.sh/uv/#highlights) has some pointers.

The PR is pretty out of date atm though, do you want me to rebase?

yakatz:
If you can rebase, that would be great. There are a bunch of other cleanup things around the build/test process that probably need to be done:

- [ ] #1302 
- [ ] Make sure tests are running properly - for example, why didn't we catch #1404 in testing?

PR #1366, opened 2026-01-11
Fix development instructions

smortex:
Running the commands from the README to start development with the
project fail.  Improve newcommers experience by updating the actual
commands to run, and give hints about how to specify a custom settings
file because one is mandatory to start the application.

### DIFF
--- a/README.md
+++ b/README.md
@@ -397,7 +397,7 @@ Write decent commit messages. Don't use swear words and refrain from uninformati
 
 The preferred format of a commit message:
 
-    docs/quickstart: Fixed a typo in the Nodes section.
+    Fixed a typo in the Nodes section.
 
     If needed, elaborate further on this commit. Feel free to write a
     complete blog post here if that helps us understand what this is


README.md
```diff
@@ -340,7 +340,7 @@ If you wish to hack on Puppetboard you should fork/clone the Github repository a
 
 ```bash
 pip install --upgrade wheel setuptools
-python setup.py develop
+pip install -e .
 pip install --upgrade -r requirements-test.txt
 mypy --install-types --non-interactive puppetboard/ test/
```
smortex: I stumble on this wanting to test a PR.  I am not a python developer, the above command allowed me to proceed, but maybe "better" options are available.   Feedback from python developers is more than welcome!

bwitt: These instructions should explain how to use a virtualenv, as it suggests just below this line.  maybe something like
```
python -m venv venv  # one-time creation of the virtual env
source venv/bin/activate
pip install --upgrade wheel setuptools pip
pip install -e .
pip install -r requirements-test.txt
mypy --install-types --non-interactive puppetboard/ test/
```

maybe we should also explain how to deactivate the virtual env?
```
deactivate
```

PR #1374, opened 2026-02-13
Add structured facts support with hierarchical display

rismoney:
Implements comprehensive support for hierarchical/structured facts (e.g., os.release.full, networking.ip) throughout the facts pages.

Features:
- Facts list page displays structured facts grouped under parent with blue headers
- Child facts shown as indented bullets under parent
- Clicking parent fact (e.g., 'os') shows full JSON structure
- Clicking child fact (e.g., 'os.release.full') shows extracted nested values
- Support for filtering by nested fact values
- Single bulk query for structure detection (avoids N+1 performance issue)

Performance: Uses single bulk PuppetDB query instead of per-fact queries for fast loading.

### DIFF
--- a/puppetboard/templates/fact.html
+++ b/puppetboard/templates/fact.html
@@ -45,7 +45,11 @@ try {
 {% if render_graph %}
 <div id="factChart" width="300" height="300"></div>
 {% endif %}
-<h1>{{ fact }}{% if value %} : <span id="value"></span>{% endif %}</h1>
+<h1>
+  {{ fact }}
+  {% if is_structured %}<span style="color: #999; font-size: 0.7em; font-weight: normal;">(structured fact)</span>{% endif %}
+  {% if value %} : <span id="value"></span>{% endif %}
+</h1>
 <table id="facts_table" class='ui fixed very basic compact table stackable'>
   <thead>
     <tr>
--- a/puppetboard/templates/facts.html
+++ b/puppetboard/templates/facts.html
@@ -1,5 +1,39 @@
 {% extends 'layout.html' %}
+
+{# Macro to render simple flat list of direct children only (no recursion) #}
+{% macro render_fact_tree(all_children, parent_path, current_env, depth) %}
+  {%- for child in all_children %}
+    <li style="font-size: 90%; list-style: none;">
+      <a href="{{url_for('fact', env=current_env, fact=child.name)}}">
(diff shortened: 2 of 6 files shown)


yakatz (commented):
I am not up to date enough with the code to give a binding reveiw, but this looks great

rismoney:
still has some bugs but its close -wanted to get feedback

rismoney:
lmk if anything should be changed.  I think its a big improvement

bwitt:
Perhaps update and/or add screenshots to the docs?

rismoney:
I had to redact facts, for privacy reasons. but os is an easy one:

image

If you click on the blue background top level structured fact, it returns the data structure, the way the facts do today.

If you click on the sub-fact, you get that 
image

bwitt:
thanks I mean are there existing screenshots in the docs that should be replaced?

rismoney:
If it is merged, oerhaps someone else should provide them with a more 'stock' environment screenshot.

rismoney:
Good thought on the memory concern. I tried to keep it pragmatic.  I have been running it for about 2 weeks.  Mount points were one area.  I ended up skipping them and a few other facts for this go round, for a few reasons.  One is usefulness, and the requirement to encode url's turned into  trickier scope creep than I wanted for little gain, based on how good the rest of it works.  I think dots are omitted too.  

So for those, at this time, you will get the same legacy json like view.   I have a ton of types of complex hashes, with variable keys aggregated and all unique keys across all nodes.  This technically can grab a chunk, but I didn't see it be problematic in my testing - I do some things with package inventory.   If you have deeply nested facts, obviously depth can result in many strings.   By lazy loading the facts, we do avoid only loading sub-facts unless clicked on.  When I started this, the page load times were so long.    I don't normally use GRAPH_FACTS with the structured fact, but it displayed too.

I felt given the project's age and new innovation coming into several alternate projects to this, the least this (and my next PR) can do, is align this with the products stance on legacy fact deprecation.

yakatz:
Works for me

yakatz:
I would like to mark this as backwards-incompatible to bump the major version due to the significance of the change and possible performance implications.

bastelfreak:
did you check if other major changes were merged already? Would a minor release prior the merge been possible?

yakatz:
> did you check if other major changes were merged already? Would a minor release prior the merge been possible?

I looked back more than 6 months and I don't see any merged PRs in that timespan except for dependabot and changes to the open shift template.

Issue #1393, opened 2026-03-12
7.0.0 structured facts not enumerating


rismoney:
I will look at this later this evening. Not sure if something changed between my commit for this or elsewhere.  Tested locally working.    I wonder if it is problematic in the container.  Let me see if I can dig in there.

yakatz:
This should be resolved. Github didn't close it automatically with the PR.

PR #1394, opened 2026-03-12
fix url prefix problem in facts.html

rismoney:
the fetch() and child used hardcoded paths that ignored PUPPETBOARD_URL_PREFIX breaking Docker container usage

### DIFF
--- a/puppetboard/core.py
+++ b/puppetboard/core.py
@@ -6,6 +6,7 @@ from importlib.metadata import version
 from flask import Flask
 from flask_caching import Cache
 from flask_apscheduler import APScheduler
+from werkzeug.middleware.proxy_fix import ProxyFix
 from pypuppetdb import connect
 
 from puppetboard.utils import (get_or_abort, jsonprint,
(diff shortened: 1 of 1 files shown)


yakatz:
Fixes #1393 ?

yakatz:
Is there any way we can test for this sort of thing? (I would guess it isn't with the effort, but have to ask )

rismoney:
yes sir

yakatz (approved):

rismoney:
I am not sure how to help testing wise, because it is client side javascript.  pytest is sort of out of scope on this, I *think*.  I probably should have revalidated the changes in the container. That would have caught it.  I didn't expect there to be a difference, but didn't dig into the docker_settings.py either.

yakatz:
Just making sure - was this change tested both in the container and out?

rismoney:
standby.

rismoney:
maybe it is working in my container all along.  i have an external issue through how i access the container and  need to see if my reverse proxy or something is messing with things.

rismoney:
It was working all along - i think it was not enumerating sub facts because of how reverse proxying works.  Found this thing I googled.  https://werkzeug.palletsprojects.com/en/stable/middleware/proxy_fix and it now works on both my local ubuntu, and a container i built locally and ran directly accessing, and through reverse proxying.  I noticed When I put my browser in F12 debug mode, I was getting json mixed mode fails in some scenarios.

Originally I had a bunch of decisions to make regarding how to load the structured facts.  Lazy loading turned out to be the most performant of the few I tried, because of how expensive it could potentially get on a large quantity of facts/data.  

So that uses a Javascript fetch().  Everything before that was standard refs loading full pages.   Browsers apparently handle these differently.  Content could be mixed (http/https) in reverse proxy situations, and this redirect thing meant it was getting blocked.

yakatz (changes requested):
Can you confirm if this is still necessary? It isn't clear from the last update.

rismoney:
yes it is still necessary

PR #1409, opened 2026-03-25
fix numerous sort issues relating to ip, disk and mem sizes

rismoney:


puppetboard/static/custom-natural-size/natural-size.js
```diff
@@ -0,0 +1,53 @@
...
+
+jQuery.extend(jQuery.fn.dataTableExt.oSort, {
+    "natural-size-pre": function (data) {
+        // DataTables strips HTML before passing to sort functions.
+        // Strip any remaining tags just in case, then trim whitespace.
+        data = String(data).replace(/<[^>]*>/g, '').trim();
+
+        var units = {
+            'b':   1,
+            'kb':  1024,
```
smortex, line 25: I guess that if you distinguish SI and IEC units, you want to use the actual values?

```suggestion
            'kb':  1000,
```

rismoney: I think facter and *most* things ae 1024 based now, regardless if they say GB or GiB.  I normalized everything for the sake of sorting here.  The thing is a lot of people mean 1024 when they say kilo in computing. My thought here  for visual presentation stuff whee a user doesn't have control over the units per se, was to align with social norms vs correctness.

PR #1411, opened 2026-03-26
proposal: change icon to 8bit vox

d1nuc0m:
More a proposal/discussion than a PR - what if we use 8bit vox as favicon? Something like [OpenVox View](https://github.com/voxpupuli/openvoxview) does.

Tagging @gdubicki as author of the actual favicon


cvquesty (approved):
LGTM

PR #1431, opened 2026-05-27
nodes: add facts timestamp column and place it last

tparkercbn:
I have servers that run puppet infrequenly but should be pushing facts every hour.  I would like to add a Facts column to the nodes screen that shows that last facts timestamp along with the last Catalog and Report timestamps.


tparkercbn:
Hello. Can this be reviewed please?

